PR #2, opened 2017-09-30
Refactor/coordinators

AlexRamey:
introduced coordinators, following pattern found here: https://github.com/wtsnz/Coordinator-Example

Note: I currently don't believe the coordinators should have a 'services' array and be able to do anything network/db related. I seem them merely as 'flow-managers' that can also interact with the store.

In general, I think the store interaction will be 
1) firing actions when their child view controllers/coordinators report navigation events to them
2) receiving actions that imply new navigation state and managing the flow accordingly


jdwitten (approved):
I'm going to approve, might add some stuff to handle routes better in the app coordinator, but its a good start-- looks good

PR #3, opened 2017-10-06
The key additions are syncAllData() and getAuthors() getCategories() and getArticles() in MBStore.swift

AlexRamey:
0. Feel free to pull down this branch and modify it yourself or to go ahead and merge it in and then to make improvements. This could use a little more refactoring. The deserialize functions in the models are a lot of copy-paste. There's probably a way to clean that up.

1. Created core data model schema and Auto-generated core data NSManagedObject subclasses
2. Added MBClient to handle HTTP details (perhaps paging logic will go here if we need to deal with that)
3. Added MBStore functionality (It is kind of doubling as a holder for our ReSwift Store object and also as a data store)
4. Exposed syncAllData() to download everything and save it into core data (updating existing objects if they exist instead of creating dups) (todo: remove old data).
5. Exposed getAuthors, getCategories, and getArticles to return arrays of these objects from CoreData.
6. Example usage code is currently in the MBArticlesViewController view did load, but this will need to get deleted since we need to invoke this functionality in the context of ReSwift newState() . . .
7. I think I'm associating authors with articles correctly, but can't tell b/c the data isn't good. None of the articles have an id of an actual author we get down (todo, speak with fishwick).
8. There is article content now though, so front end work can commence . . .

9. The print statements should notify you that everything is working out of the box. Pull down and let me know if there are any errors


iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -17,7 +17,40 @@ class MBArticlesViewController: UIViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        // Do any additional setup after loading the view, typically from a nib.
+        /******** START EXAMPLE ***********/
```
AlexRamey, line 20: this is example code so u can quickly see everything working when you spin this project up

iOS Client/Models/MBArticle+Custom.swift
```diff
@@ -0,0 +1,89 @@
...
+        
+        // TODO -- Link Categories --
+        // . . .
+        
+        return nil
+    }
+    
+}
+
+//@NSManaged public var content: String?
```
AlexRamey: need to delete this commented out stuff down here . . .

iOS Client/Models/MBDeserializable.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Copyright © 2017 Mockingbird. All rights reserved.
+//
+
+import CoreData
+
+protocol MBDeserializable {
+    static func deserialize(json: NSDictionary, intoContext managedContext: NSManagedObjectContext) -> Error?
+}
+
+//protocol ArticleLoadState {
```
AlexRamey: need to delete this comment (and maybe this entire file, I don't think we need it . . .)

iOS Client/MBClient.swift
```diff
@@ -0,0 +1,181 @@
...
+            return
+        }
+        
+        print("firing initial getAuthors request")
+        self.session.dataTask(with: url) { (data: Data?, resp: URLResponse?, err: Error?) in
+            self.pagingHandler(url: urlString, data: data, resp: resp, err: err, completion: completion)
+        }.resume()
+    }
+    
+    private func pagingHandler(url: String, data: Data?, resp: URLResponse?, err: Error?, completion: @escaping ([Data], Error?) -> Void) {
```
jdwitten, line 109: I think this is fine for now but in the future we may want to implement some logic on when to fetch new pages. It could be a little risky to just continually fire off fetch requests if there are hundreds of pages. I think this can come in when we start looking at front end stuff, maybe when the user has scrolled a certain length through the content then fire off more requests.

jdwitten (approved):

iOS Client/MBClient.swift
AlexRamey: agreed. So right now, I think we have to get all authors and all categories unless we want to fetch categories/authors individually on demand/lazily (which might be a better way). Just b/c some article could have an author that comes back on the last page of results.

For articles right now, it just gets the most recent 25. I think we need some paging solution like you're talking about for getting articles so we can dynamically fetch more as they scroll.

PR #4, opened 2017-10-10
Refactor/routing

jdwitten:
Main Idea:
- Introduction of a Route where every tab has a hierarchy of views saved in the state.
- As we learn more about the types of Routes we need, we can add to they route types (for now I just have .base and .detail where base is the main view of the tab and detail is a view containing more info on one item in a table view/ collection view in the base view).
- Coordinators subscribe to the state and manage changes in the routes that are applicable to them.
- Maintain a selectedTab in the state 
- What you gain from this is the ability to change Navigation state all through actions, so you can simulate navigating to a specific page in the app via Actions. This can be useful for push notifications.

- I'm going to throw together some front end stuff cause it looks like you nailed the Core Data/ fetching syncing stuff already


iOS Client/ArticlesController/ArticlesCoordinator.swift
```diff
@@ -24,5 +27,15 @@ class ArticlesCoordinator: NSObject, Coordinator {
     func start() {
         let articlesController = MBArticlesViewController.instantiateFromStoryboard()
         self.navigationController.pushViewController(articlesController, animated: true)
+        MBStore.sharedStore.subscribe(self)
+    }
+    
+    // MARK: - StoreSubscriber
+    func newState(state: MBAppState) {
+        guard state.navigationState.selectedTab != .articles, let newRoute = state.navigationState.routes[.articles] else {
```
AlexRamey: nice chained condition 👍

AlexRamey: shouldn't this be `state.navigationState.selectedTab == .articles`?

iOS Client/Coordinator/Coordinator.swift
```diff
@@ -24,4 +27,28 @@ extension Coordinator {
...
+    func build(newRoute: [Route]) {
+        guard let root = rootViewController as? UINavigationController else { return }
+        
+        var rootViewControllers = root.viewControllers
+        for newRouteIndex in 0..<newRoute.count {
+            if newRouteIndex > route.count - 1 { //case: more controllers in new route so push onto nav stack
+                root.pushViewController(Route.viewController(forRoute: newRoute[newRouteIndex], inTab: Tab.tab(forCoordinator: self)!)!, animated: true)
+                
+            } else if newRoute[newRouteIndex] != route[newRouteIndex] { //case: differing routes so replace top of stack with new route
+                let newTopOfStack = newRoute[newRouteIndex...newRoute.count].flatMap { Route.viewController(forRoute: $0, inTab: Tab.tab(forCoordinator: self)!)}
```
AlexRamey: flatMap with swift is something I feel like I've never fully understood. Why use a flatMap here rather than a plain map? Is the reasoning to strip out nil values so we are only pushing legit VCs?

iOS Client/Coordinator/Coordinator.swift
```diff
@@ -24,4 +27,28 @@ extension Coordinator {
...
+                rootViewControllers.removeLast(route.count - newRouteIndex)
+                newTopOfStack.forEach {
```
AlexRamey: oooooo, this forEach stuff is like javascript. Swift is becoming a juggernaut 🐘

AlexRamey: I've been using map a lot for this and ignoring the results, forEach is nicer . . .

iOS Client/Coordinator/Coordinator.swift
```diff
@@ -24,4 +27,28 @@ extension Coordinator {
...
+                    root.pushViewController($0, animated: true)
+                }
+                break
+            }
+        }
+        if newRoute.count < rootViewControllers.count { //case: less controllers in new route so pop off existing routes
```
AlexRamey: I guess this is equivalent, but `newRoute.count < route.count` would be more consistent

iOS Client/Coordinator/Coordinator.swift
```diff
@@ -24,4 +27,28 @@ extension Coordinator {
...
+            rootViewControllers.removeLast(route.count - rootViewControllers.count)
```
AlexRamey: shouldn't this be `rootViewControllers.removeLast(route.count - newRoute.count)`. I think this will always be 0 (route.count should theoretically be the same as rootViewControllers.count)

iOS Client/Coordinator/Coordinator.swift
AlexRamey: maybe more efficient to replace VCs individually as needed instead of overwriting the whole stack, but this is fine as is for now

AlexRamey: I think you can omit newRoute.count and just do `newRoute[newRouteIndex...]`

iOS Client/Coordinator/Coordinator.swift
```diff
@@ -24,4 +27,28 @@ extension Coordinator {
...
+            if newRouteIndex > route.count - 1 { //case: more controllers in new route so push onto nav stack
+                root.pushViewController(Route.viewController(forRoute: newRoute[newRouteIndex], inTab: Tab.tab(forCoordinator: self)!)!, animated: true)
+                
+            } else if newRoute[newRouteIndex] != route[newRouteIndex] { //case: differing routes so replace top of stack with new route
+                let newTopOfStack = newRoute[newRouteIndex...newRoute.count].flatMap { Route.viewController(forRoute: $0, inTab: Tab.tab(forCoordinator: self)!)}
+                rootViewControllers.removeLast(route.count - newRouteIndex)
+                newTopOfStack.forEach {
+                    root.pushViewController($0, animated: true)
+                }
+                break
```
AlexRamey, line 45: could use a `return` here, but it doesn't matter b/c the following condition `if newRoute.count < rootViewControllers.count` would never be true after this code runs.

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -33,32 +37,31 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate {
...
             return coord.rootViewController
         })
     }
     
+    // MARK: - StoreSubscriber
+    
+    func newState(state: MBAppState) {
+        if state.navigationState.selectedTab != selectedTab, let rootVC = rootViewController as? MBTabBarController {
+            rootVC.select(tab: state.navigationState.selectedTab)
+            selectedTab = state.navigationState.selectedTab
```
AlexRamey: could switch the order here and go 
```
selectedTab = state.navigationState.selectedTab
rootVC.select(tab: selectedTab)
```

iOS Client/Routing/Routing.swift
```diff
@@ -0,0 +1,82 @@
...
+                if let vc = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "BookmarksController") as? MBBookmarksViewController {
+                    return vc
+                }
+            }
+        }
+        return nil
+    }
+}
+
+enum Tab: String {
```
AlexRamey: any reason we need `String` here? If we don't, we could do `enum Tab: Int` and use `rawValue` instead of needing the computed index property below.

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -21,24 +21,19 @@ class MBTabBarController: UITabBarController, StoreSubscriber {
...
     }
     
-    // MARK: - StoreSubscriber
-    
-    func newState(state: MBAppState) {
-        if self.selectedIndex != state.navigationState.tabIndex {
-            self.selectedIndex = state.navigationState.tabIndex
-        }
+    func select(tab: Tab) {
+        self.selectedIndex = tab.index
```
AlexRamey: investigate going `Tab: Int` and using `rawValue`. I might be missing something about why we would need String to be the raw value type . . .

AlexRamey (changes requested):
very nice, learned a lot looking through this. I think routes are cool. I need to read up on them for some more context b/c it's new to me. Most of my feedback is micro-stuff, the setup seems solid from what I can tell.

iOS Client/Coordinator/Coordinator.swift
jdwitten: yeah so flatMap will discard null values for you, but it also treats nested lists as one big long list. So [[1,2,3,4], [5,6,7,8], [9,10,11,12]] is treated as [1,2,3,4,5,6,7,8,9,10,11,12]. So it basically flattens all nested lists. I'm just using it here to make sure there are no null values

iOS Client/TabController/MBTabBarController.swift
jdwitten: I like this better, good call

jdwitten:
Yo check out the changes I made. You had some good points and I addressed pretty much all of them.

iOS Client/Coordinator/Coordinator.swift
AlexRamey: nice

AlexRamey (approved):
money!

PR #5, opened 2017-10-13
Feature/display articles

jdwitten:
- I had to do a little bit of funky stuff with the back button on the article detail to get it to play nicely with the coordinators. We can revisit this later to maybe do something a little less hacky.
- Web view turned out nice it handles all images for us and even as a nice player for videos
- Definitely quick and dirty but it will get the job done for a quick demo with DZ tomorrow


AlexRamey (approved):
looks good, I'll go over this again this weekend more thoroughly but I like what I see after a cursory glance

PR #6, opened 2017-10-15
Background App Refresh (plus debug utility)

AlexRamey:
This PR is starting down the road of first trying to read from the device for article data when the app launches.

If there is no data or it is over a week old, then a network sync is attempted.

Network syncs are attempted periodically via Background App Refresh and timestamps are stored in UserDefaults to record the last successful sync. 

There is a debug utility right now in the nav bar that shows you the last time any sync was successful and the last time a background app refresh sync was successful.

Also had to make some changes to the data loading code to pass back if new data was found, in order to report it to Background App Refresh.


iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -34,60 +35,51 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
+        guard let managedContext = appDelegate.persistentContainer?.viewContext else {
+            print("Unable to get the managed object context!")
+            return
+        }
+  
+        
+        let oneWeekAgoTimestamp = Date().timeIntervalSinceReferenceDate - MBConstants.SECONDS_IN_A_WEEK
+        let lastUpdateTimestamp = UserDefaults.standard.double(forKey: MBConstants.DEFAULTS_KEY_ARTICLE_UPDATE_TIMESTAMP)
+        
+        if lastUpdateTimestamp > oneWeekAgoTimestamp {
```
jdwitten, line 57: This is where I feel like we don't necessarily need to be doing this kind of update in the view controller. I don't know exactly where to put this, maybe in the App Coordinator instead in its start method. I think it's ok for now because we are still figuring out this syncing stuff but down the road might want to change it

jdwitten (approved):

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: agreed. I will clean this up once I solve the random crashing issue

PR #7, opened 2017-10-21
Refactoring/Cleaning up code. Added CSS for improved reading experience.

AlexRamey:
We are in CSS-land now! The question is this: How to get the Youtube videos to render . . . We need to experiment around. This could be tough based on my initial research . . .


iOS Client/Actions/MBActions.swift
```diff
@@ -13,7 +13,7 @@ struct NavigationActionSwitchTab: Action {
 }
 
 struct PushRoute: Action {
-    var route: Route
+    var route: RouteComponent
```
jdwitten, line 16: I like this name better-- nice

iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -44,29 +60,8 @@ class MBArticleDetailViewController: UIViewController, StoreSubscriber, WKUIDele
         MBStore.sharedStore.dispatch(PopCurrentNavigation())
     }
     
-    func newState(state: MBAppState) {
```
jdwitten, line 47: Does this work without subscribing to the store? How does the VC know what article to display?

jdwitten (changes requested):

iOS Client/ArticlesController/MBArticleDetailViewController.swift
AlexRamey: Beforehand, this was working, but it kind of sketched me out b/c this was the flow:
1) User selects article in table view
2) table view controller fires selected article action
3) reducer processes action and updates state
4) then the article coordinator would get the new state and create the detail view controller, which would then itself subscribe to the store
5) Then, (even though no further action was fired) the detail view controller would get notified of the new state and set up its web view based on the article in the state

I was worried that sometimes this new detail view controller might not get notified of the state upon creation since no new action is actually fired after it is created. Furthermore, I thought it would just be simpler for the coordinator to inject the selected article into this view controller upon creation. So now you'll notice that `instantiateFromStoryboard()` in this class accepts the selected article and sets it on the returned detail view controller.

iOS Client/Routing/Routing.swift
```diff
@@ -25,54 +24,28 @@ enum Route: Equatable {
...
-                    return vc
-                }
-            } else if case .detail = route {
-                if let vc = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "ArticleDetailController") as? MBArticleDetailViewController {
-                    return vc
-                }
+            if case .base = self {
+                retVal = MBArticlesViewController.instantiateFromStoryboard()
+            } else if case let .detail(detailItem) = self {
+                retVal = MBArticleDetailViewController.instantiateFromStoryboard(article: detailItem as? MBArticle)
```
AlexRamey, line 35: here is where the article gets passed in

iOS Client/Routing/Routing.swift
```diff
@@ -25,54 +24,28 @@ enum Route: Equatable {
     }
  
     
-    static func viewController(forRoute route: Route, inTab tab: Tab) -> UIViewController? {
+    func viewController(forTab tab: Tab) -> UIViewController? {
```
AlexRamey, line 27: and this is now an instance method on route component enum, so the article can be accessed from the associated value in case where self is `detail`. Thought this would be a little cleaner, but up for other opinions here . . .

iOS Client/ArticlesController/MBArticleDetailViewController.swift
jdwitten: Ok I see,  I like it I will approve

jdwitten (approved):

PR #8, opened 2017-11-02
Backend/image data

AlexRamey:
I added `imageID` integer field to the existing article model (gets set on data sync).

I also provided a method called `getImageData` on the MBClient class. It accepts a closure that takes an optional UIImage. I think we can rely on the OS to cache these images for us, but I can implement some caching in the documents directory if we need performance.

EXAMPLE (Not checked in):

```swift
class MBBookmarksViewController: UIViewController, StoreSubscriber {
    @IBOutlet weak var imageView : UIImageView!
    private let client: MBClient = MBClient()
    static func instantiateFromStoryboard() -> MBBookmarksViewController {
        // swiftlint:disable force_cast
        return UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "BookmarksController") as! MBBookmarksViewController
        // swiftlint:enable force_cast
    }

    override func viewDidLoad() {
        super.viewDidLoad()
        MBStore.sharedStore.subscribe(self)
        // Do any additional setup after loading the view.
    }
    
        switch state.articleState.articles {
        case .initial:
        break //Do nothing if articles haven't tried to load
        case .loading:
        break //Do something here to indicate loading
        case .error:
            print("Error: Loading articles")
        case .loaded(let data):
            if data.count > 0 {
                client.getImageData(imageID: Int(data[0].imageID), completion: { (image: UIImage?) in
                    if image != nil {
                        print("money")
                        DispatchQueue.main.async {
                            self.imageView.image = image
                        }
                    } else {
                        print("anti money!")
                    }
                })
            }
        }
    }
```


.gitignore
```diff
@@ -21,6 +21,7 @@ xcuserdata/
 *.moved-aside
 *.xccheckout
 *.xcscmblueprint
+*.DS_Store
```
jdwitten, line 24: good call I have just been discarding changes

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -106,6 +108,62 @@ class MBClient: NSObject {
         }.resume()
     }
     
+    func getImageData(imageID: Int, completion: @escaping (UIImage?) -> Void) {
```
jdwitten, line 111: Nice looks good

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -106,6 +108,62 @@ class MBClient: NSObject {
...
+                    
+                    completion(image)
+                }.resume()
+            } else {
+                completion(nil)
+            }
+            
+        }.resume()
+    }
+
```
jdwitten, line 166: I'm wondering if we should have used a library like PromiseKit so we don't get nested completion blocks like this. Might be too late but it has been pretty helpful on Synchrony.

jdwitten (approved):

PR #10, opened 2017-11-04
Feature/devotions

jdwitten:
- Added Devotions tab, coordinator, view controller
- I'm just reading from a json file at this point, we can change the client function to go over the network at some point
- I'll make another PR for notifications, this one was getting big as is.


iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -0,0 +1,46 @@
...
+    }()
+    
+    func start() {
+        let devotionsController = MBDevotionsViewController.instantiateFromStoryboard()
+        navigationController.pushViewController(devotionsController, animated: true)
+        MBStore.sharedStore.subscribe(self)
+        
+        MBStore().syncDevotions { devotions, error in
+            if error != nil {
+                MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.error))
```
AlexRamey, line 31: Do these dispatches in a `DispatchQueue.main.async` block. I read on the ReSwift repo:

```
Note that the callback block from the network request arrives on a background thread, therefore we're using dispatch_async(dispatch_get_main_queue()) to perform the state update on the main thread. ReSwift will call reducers and subscribers on whatever thread you have dispatched an action from. We recommend to always dispatch from the main thread, but ReSwift does not enforce this recommendation. ReSwift will enforce that all Dispatches, Store Subscribes and Store Unsubscribes are on the same thread or serial Grand Central Dispatch queue. Therefore the main dispatch queue works, however the global dispatch queue, being concurrent, will fail.
```

AlexRamey: For now, it's fine b/c all the code in syncDevotions and getDevotionsWithCompletion is currently synchronous.

AlexRamey (dismissed):

AlexRamey (approved):
nice

PR #19, opened 2017-11-15
1. Group articles by category in the table view

AlexRamey:
I left comments throughout to annotate things.

The big pieces of this PR are:
1) Added `parent` relationship from category to category (one to one)
2) Link up the categories after downloading them all (set parent to point to the correct parent)
3) Helper instance method on MBCategory to get its Top-Level Category (walk up the tree to a root)
4) ArticlesViewController uses this infrastructure to group articles by top-level category name
5) bug fixes in http client (I was doing paging wrong)
6) Custom section header views to label each section
7) Tweaks to custom cell to make it display nicer for now
8) Removed debug utility

### DIFF
--- a/iOS Client/ArticlesController/ArticleTableViewCell.swift
+++ b/iOS Client/ArticlesController/ArticleTableViewCell.swift
@@ -38,12 +38,13 @@ class ArticleTableViewCell: UITableViewCell {
                 }
             }
         }
+        self.coverImage.image = nil
         titleLabel.attributedText = title
         titleLabel.font = UIFont.systemFont(ofSize: 18)
         titleLabel.textColor = UIColor.ArticleTitle
         authorLabel.attributedText = author
         authorLabel.font = UIFont.boldSystemFont(ofSize: 15)
-        snippetLabel.text = snippet
+        snippetLabel.text = ""//snippet
     }
     
 }
--- a/iOS Client/ArticlesController/MBArticlesViewController.swift
+++ b/iOS Client/ArticlesController/MBArticlesViewController.swift
@@ -12,9 +12,8 @@ import CoreData
 
 class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableViewDataSource, StoreSubscriber {
     @IBOutlet weak var tableView: UITableView!
-    var articles: [MBArticle] = []
-    var attributedTitles: [NSAttributedString] = []
-    var attributedAuthors: [NSAttributedString] = []
+    var articlesByCategory = [String: [MBArticle]]()
+    var topLevelCategories: [String] = []
     let client = MBClient()
     static let ArticleTableViewCellId = "ArticleTableViewCell"
     
@@ -42,11 +41,6 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
         tableView.register(UINib(nibName: MBArticlesViewController.ArticleTableViewCellId, bundle: nil), forCellReuseIdentifier: MBArticlesViewController.ArticleTableViewCellId)
         tableView.rowHeight = UITableViewAutomaticDimension
         tableView.estimatedRowHeight = CGFloat(MBConstants.ARTICLE_TABLEVIEWCELL_HEIGHT)
-        
-        
-        // Set up a bar button item to toggle debug info on background app refresh
-        let item = UIBarButtonItem(barButtonSystemItem: .refresh, target: self, action: #selector(MBArticlesViewController.showTimestamps))
-        self.navigationItem.setRightBarButton(item, animated: false)
         
         // 1. the managed context has to be passed in (UIApplication should only be accessed from main thread)
         guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else {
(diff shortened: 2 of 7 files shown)


iOS Client/ArticlesController/ArticleTableViewCell.swift
```diff
@@ -38,12 +38,13 @@ class ArticleTableViewCell: UITableViewCell {
...
             }
         }
+        self.coverImage.image = nil
         titleLabel.attributedText = title
         titleLabel.font = UIFont.systemFont(ofSize: 18)
         titleLabel.textColor = UIColor.ArticleTitle
         authorLabel.attributedText = author
         authorLabel.font = UIFont.boldSystemFont(ofSize: 15)
-        snippetLabel.text = snippet
+        snippetLabel.text = ""//snippet
```
AlexRamey, line 47: I'm hiding this for now so it looks better

iOS Client/ArticlesController/ArticleTableViewCell.swift
```diff
@@ -38,12 +38,13 @@ class ArticleTableViewCell: UITableViewCell {
                 }
             }
         }
+        self.coverImage.image = nil
```
AlexRamey, line 41: this prevents the wrong image temporarily being on the cells as you scroll

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -43,11 +42,6 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
         tableView.rowHeight = UITableViewAutomaticDimension
         tableView.estimatedRowHeight = CGFloat(MBConstants.ARTICLE_TABLEVIEWCELL_HEIGHT)
         
-        
-        // Set up a bar button item to toggle debug info on background app refresh
-        let item = UIBarButtonItem(barButtonSystemItem: .refresh, target: self, action: #selector(MBArticlesViewController.showTimestamps))
-        self.navigationItem.setRightBarButton(item, animated: false)
-
```
AlexRamey, line 50: I removed this debug utility

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -109,73 +103,84 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
         case .error:
             print("Error: Loading articles")
         case .loaded(let data):
-            if data.count != articles.count {
```
AlexRamey, line 112: took out this condition since the count doesn't necessarily tell us whether or not the data is new.

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -12,9 +12,8 @@ import CoreData
 
 class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableViewDataSource, StoreSubscriber {
     @IBOutlet weak var tableView: UITableView!
-    var articles: [MBArticle] = []
-    var attributedTitles: [NSAttributedString] = []
```
AlexRamey, line 16: I just compute these on the fly now in the call to configure. Didn't want to deal with a data structure for these since articles is no longer a simple list. Also didn't want to add an extension on the model objects for fear of core data (but maybe that'd be fine)

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -109,73 +103,84 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
     }
     
     override func didReceiveMemoryWarning() {
         super.didReceiveMemoryWarning()
         // Dispose of any resources that can be recreated.
     }
-    
-    // MARK: - showTimestamps() is used for debug purposes
-    @objc func showTimestamps() {
-        let lastOverallUpdate = UserDefaults.standard.double(forKey: MBConstants.DEFAULTS_KEY_ARTICLE_UPDATE_TIMESTAMP)
```
AlexRamey, line 129: removed this debug utility

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -109,73 +103,84 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
     }
+    
+    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        let view = UIView(frame: CGRect(x: 0, y: 0, width: tableView.frame.size.width, height: 30))
+        let label = UILabel(frame: CGRect(x: 0, y: 0, width: tableView.frame.size.width, height: 30))
+        label.text = topLevelCategories[section]
+        view.addSubview(label)
+        view.backgroundColor = UIColor.orange
+        return view
+    }
```
AlexRamey, line 184: temporary section header view. This can be spiffied up later

iOS Client/Models/MBCategory+Custom.swift
```diff
@@ -48,9 +48,20 @@ extension MBCategory {
...
+    // This method follows the parent links until it encounters a top-level category,
+    // which it returns. If called on a top-level category, this method returns the
+    // receiver.
+    public func getTopLevelCategory() -> MBCategory? {
+        if self.parentID == 0 {
+            return self
+        } else {
+            return self.parent?.getTopLevelCategory() ?? nil
+        }
+    }
```
AlexRamey, line 66: recursion is straight cash for climbing up trees

iOS Client/Store/MBStore.swift
```diff
@@ -61,13 +61,20 @@ class MBStore: NSObject {
...
                     } else {
-                        self.downloadArticles(persistentContainer: persistentContainer, completion: { (isNewArticleData, articleErr) in
-                            isNewData = isNewAuthorData ?? false || isNewCategoryData ?? false || isNewArticleData ?? false
-                            if let unwrappedArticleErr = articleErr {
-                                print("There was an error downloading article data! \(unwrappedArticleErr)")
-                                completion(nil, articleErr)
+                        self.linkCategoriesTogether(persistentContainer: persistentContainer, completion: { (linkErr) in
+                            if let unwrappedLinkErr = linkErr {
+                                print("There was an error linking categories! \(unwrappedLinkErr)")
+                                completion(nil, linkErr)
```
AlexRamey, line 67: so now, during the data sync, after we get down all the categories, we have to go through them and set the parent relationship to point to the correct parent category (or nil if parentID is 0)

iOS Client/ArticlesController/ArticleTableViewCell.swift
jdwitten: Yeah that was more of an experiment you can do whatever you want with it

iOS Client/ArticlesController/MBArticlesViewController.swift
jdwitten: Yeah this is actually a pretty big topic of discussion on how to tell if you have new data. There are whole libraries to do table diffing and such. We can probably just get by with reloading everytime and if it seems to impact performance then we address it

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -109,73 +103,84 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
+            
+                for topLevelCategory in topLevelCategories {
+                    if let _ = retVal[topLevelCategory] {
+                        retVal[topLevelCategory]?.append(article)
+                    } else {
+                        retVal[topLevelCategory] = [article]
+                    }
+                }
+                
+                if topLevelCategories.count == 0 {
```
jdwitten, line 130: We might still want to display these articles some how. Not sure exactly where, probably depends on how the design ends up

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -109,73 +103,84 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
-        self.present(ac, animated: true, completion: nil)
-    }
 }
 
-// MARK: - UITableViewDataSource
 extension MBArticlesViewController {
+    // MARK: - UITableViewDataSource
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return articles.count
+        return articlesByCategory[topLevelCategories[section]]!.count
```
jdwitten, line 149: Use an optional here instead of a ! to unwrap 
`articlesByCategory[topLevelCategories[section]]?.count ?? 0`

or something of the sort so we don't crash just because we have no articles

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -109,73 +103,84 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
     }
     
     func numberOfSections(in tableView: UITableView) -> Int {
-        return 1
+        return topLevelCategories.count
     }
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let article = articles[indexPath.row]
+        let article = articlesByCategory[topLevelCategories[indexPath.section]]![indexPath.row]
```
jdwitten, line 157: I would say the same here try not to use the !

jdwitten (dismissed):
Looks good, just a few small things about unwrapping optionals

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: I could do this here pretty cleanly, but I don't want to deal with this case in cellForRowAt and didSelectRowAt.

My argument is that this is a good place to use the `!` b/c it is a programmer error and we want this to crash if our assumptions are wrong so we can fix them.

lines 107,108 from above:
```
articlesByCategory = groupArticlesByTopLevelCategoryName(articles: data)
topLevelCategories = Array(articlesByCategory.keys)
```

we know that topLevelCategories only contains keys of articlesByCategory, so we know that the result of a key lookup will not be nil

jdwitten: Yeah that's a good call it should never really be nil in the wild. I'm going to approve.

jdwitten (approved):

PR #20, opened 2017-11-21
Feature/devotion notifications

jdwitten:
* Creates a Devotion Detail Controller (all it is is a label with the content on it)
* Prompt the user for notifications and schedule notifications if granted
* Handle notifications when the user opens the app and take them straight to the devotion page with the devotion for that day


iOS Client/Store/MBStore.swift
```diff
@@ -22,6 +22,21 @@ class MBStore: NSObject {
...
+    static func loadState() -> MBAppState {
+        let defaults = UserDefaults.standard
+        var notificationPermission: Permission = .unprompted
+        if let notificationSetting = defaults.object(forKey: "notifications") as? Bool {
+            notificationPermission = notificationSetting ? .approved : .denied
+        }
+        return MBAppState(navigationState: MBNavigationState(), articleState: MBArticleState(), devotionState: MBDevotionState(), settingsState: MBSettingsState(notificationPermission: notificationPermission))
+    }
+    
+    static func saveState(state: AppState) {
```
AlexRamey, line 35: where are these functions being called?

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
             self.addChildCoordinator(childCoordinator: coord)
             return coord.rootViewController
         })
+        promptForNotifications()
```
AlexRamey, line 47: This stuff belongs in the devotions coordinator i think

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
+        devotions.forEach { devotion in
+            guard let devotionDay = Formatters.devotionDateFormatter.date(from: devotion.date), let calendar = Formatters.calendar else {
+                return
+            }
+            var dateComponents = DateComponents()
+            dateComponents.hour = 10
+            dateComponents.minute = 30
+            dateComponents.year = calendar.component(NSCalendar.Unit.year, from: devotionDay)
+            dateComponents.month = calendar.component(NSCalendar.Unit.month, from: devotionDay)
+            dateComponents.day = calendar.component(NSCalendar.Unit.day, from: devotionDay)
```
AlexRamey, line 83: let's move all this logic to a derived property on MBDevotion class. It could be an optional property of type DateComponents called dateComponents or something, and then here we could just access that lazy property and continue if it's nil or keep going if it's not.

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
+        }
+    }
+    
+    func scheduleNotifications(devotions: [MBDevotion]) {
+        print("scheduling notifications")
+        let center = UNUserNotificationCenter.current()
+        center.removeAllPendingNotificationRequests()
+        devotions.forEach { devotion in
+            guard let devotionDay = Formatters.devotionDateFormatter.date(from: devotion.date), let calendar = Formatters.calendar else {
+                return
```
AlexRamey, line 76: probably want `continue` so we don't miss out on the ones after this.

Honestly, might be worth crashing if there's something wrong, especially if we just bundle in the devotional data b/c then we can catch any issues before release

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
+            var dateComponents = DateComponents()
+            dateComponents.hour = 10
+            dateComponents.minute = 30
+            dateComponents.year = calendar.component(NSCalendar.Unit.year, from: devotionDay)
+            dateComponents.month = calendar.component(NSCalendar.Unit.month, from: devotionDay)
+            dateComponents.day = calendar.component(NSCalendar.Unit.day, from: devotionDay)
+            let trigger = UNCalendarNotificationTrigger(dateMatching: dateComponents, repeats: true)
+            let content = UNMutableNotificationContent()
+            content.title = devotion.verse
+            content.body = "Read your daily devotion from Mockingbird"
```
AlexRamey, line 87: lol do we include a random emoji every day? 💥🎉🙃

AlexRamey: It might be worth subclassing UNMutableNotificationContent and setting up these default values in an `init`

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -62,4 +108,19 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
         
         return false
     }
+    
+    // MARK: - UNNotificationDelegate
+    func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
```
AlexRamey, line 113: this should probably go in the devotions coordinator as well.

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -62,4 +108,19 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
+        MBStore().syncDevotions { devotions, error in
+            if error != nil {
+                print("Could not load devotion from notifications")
+            } else {
+                let date = Date()
+                if let devotion = devotions?.first(where: { Formatters.devotionDateFormatter.date(from: $0.date) == date}) {
```
AlexRamey, line 119: maybe worth putting these in core data later, but this is fine for now

iOS Client/Utilities/Formatters.swift
```diff
@@ -0,0 +1,22 @@
...
+    static let devotionDateFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-DD"
+        return formatter
+    }()
+    
+    static let calendar: NSCalendar? = {
+       return NSCalendar.init(calendarIdentifier: NSCalendar.Identifier.gregorian)
+    }()
+}
```
AlexRamey, line 22: If we compute the DateComponents as a lazy property in the MBDevotions class, then we could move these details into that class as well, either as private helper methods or private lazy properties, since they are specific to the MBDevotion objects.

iOS Client/TabController/AppCoordinator.swift
AlexRamey: we need to move a lot of this boiler plate out of this function

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
         })
+        promptForNotifications()
+    }
+    
+    func promptForNotifications() {
+        let center = UNUserNotificationCenter.current()
+        
+        center.requestAuthorization(options: [.alert, .badge, .sound]) { granted, error in
+            if granted {
+                MBStore().syncDevotions { devotions, error in
```
AlexRamey, line 55: New thought: How bout, in the DevotionCoordinator `newState`, we check to see if there are devotions. If there are, then we call promptForNotifications (and pass the devotions to it as a parameter), and then it simply asks the user for permission, and, if granted, calls scheduleNotifications on the devotions?

iOS Client/Store/MBStore.swift
```diff
@@ -22,6 +22,21 @@ class MBStore: NSObject {
         super.init()
     }
     
+    
+    static func loadState() -> MBAppState {
```
AlexRamey, line 26: where is this called?

AlexRamey (changes requested):
functionality looks good, we just need to refine the code structure a little more

iOS Client/TabController/AppCoordinator.swift
AlexRamey: and `newState` or `promptForNotifications` could be smart about checking if the user has already been prompted or not?

iOS Client/Store/MBStore.swift
jdwitten: These were just an idea I had to save and load the state. I didn't actually get around to using them so I will remove for now

AlexRamey:
these changes were included by #21

PR #21, opened 2017-11-29
Feature/devotion detail

jdwitten:
- This gets us started with a devotion detail design and a custom devotion detail table view cell.
- Also, records if a devotion has been read and displays a dot if it is unread on the table view


iOS Client/DevotionsController/DevotionDetailViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+        MBStore.sharedStore.unsubscribe(self)
+    }
+
+    func newState(state: MBAppState) {
+        if let devotion = state.devotionState.selectedDevotion,
+            let date = Formatters.devotionDateFormatter.date(from: devotion.date),
+            let monthInt = Formatters.calendar?.component(.month, from: date),
+            let day = Formatters.calendar?.component(.day, from: date) {
+            monthLabel.text = Formatters.getMonth(fromInt: monthInt)
+            dayLabel.text = String(describing: day)
```
AlexRamey: that's cool, never seen String describing before

iOS Client/DevotionsController/DevotionDetailViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+    @IBOutlet weak var monthLabel: UILabel!
+    @IBOutlet weak var dayLabel: UILabel!
+    @IBOutlet weak var verseLabel: UILabel!
+    @IBOutlet weak var verseTextLabel: UILabel!
+    @IBOutlet weak var bodyLabel: UILabel!
+    
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        configureBackButton()
```
AlexRamey, line 23: maybe we should subclass UIViewController and override viewDidLoad to configure the back button for us since we do it more than once? We could also do the store subscribe and unsubscribe in viewWillAppear and viewWillDisappear since we do that in all our VCs. Thinking out loud here, I can handle it in a later refactor if u think it's a good idea

iOS Client/DevotionsController/DevotionDetailViewController.swift
AlexRamey: maybe worth saving the formatted month string and the formatted day string as lazy properties on the devotion class so we don't have to burden this function with these details

iOS Client/DevotionsController/DevotionTableViewCell.swift
```diff
@@ -0,0 +1,28 @@
...
+    @IBOutlet weak var verseLabel: UILabel!
+    @IBOutlet weak var readNotifier: UIImageView!
+    
+    
+    func configure(day: String, month: String, verse: String, read: Bool) {
+        dayLabel.text = day
+        monthLabel.text = month
+        verseLabel.text = verse
+        
+        readNotifier.layer.cornerRadius = readNotifier.bounds.width / 2
```
AlexRamey: very much looking forward to seeing this in action

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -12,11 +12,14 @@ import ReSwift
...
     @IBOutlet weak var tableView: UITableView!
-    var devotions: [MBDevotion] = []
+    var devotions: [LoadedDevotion] = []
     override func viewDidLoad() {
         super.viewDidLoad()
         tableView.delegate = self
         tableView.dataSource = self
+        tableView.rowHeight = UITableViewAutomaticDimension
+        
+        tableView.register(UINib(nibName: "DevotionTableViewCell", bundle: nil), forCellReuseIdentifier: "DevotionTableViewCell")
```
AlexRamey: it would be good to make a constant called `reuseIdentifier` as a private property in this class

AlexRamey: and use it here and in tableView cellForRowAt

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -52,8 +55,15 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let devotion = devotions[indexPath.row]
-        let cell = tableView.dequeueReusableCell(withIdentifier: "DevotionTableViewCell") ?? UITableViewCell()
-        cell.textLabel?.text = devotion.text
+        //swiftlint:disable force_cast
+        let cell = tableView.dequeueReusableCell(withIdentifier: "DevotionTableViewCell") as! DevotionTableViewCell
+        //swiftlint:enable force_cast
+        if let date = Formatters.devotionDateFormatter.date(from: devotion.devotion.date),
+            let day = Formatters.calendar?.components(.day, from: date).day,
+            let monthInt = Formatters.calendar?.component(.month, from: date) {
```
AlexRamey: if we move this into the devotion class as a lazy property, we don't have to repeat these computations everywhere. Maybe they will have to be optional since there's a chance of failure, or maybe we should throw in `!` and fail fast since we theoretically control the data source

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -62,4 +108,19 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
         
         return false
     }
+    
+    // MARK: - UNNotificationDelegate
+    func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
```
AlexRamey: this should be devotion coordinator's concern

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
+        devotions.forEach { devotion in
+            guard let devotionDay = Formatters.devotionDateFormatter.date(from: devotion.date), let calendar = Formatters.calendar else {
+                return
+            }
+            var dateComponents = DateComponents()
+            dateComponents.hour = 10
+            dateComponents.minute = 30
+            dateComponents.year = calendar.component(NSCalendar.Unit.year, from: devotionDay)
+            dateComponents.month = calendar.component(NSCalendar.Unit.month, from: devotionDay)
+            dateComponents.day = calendar.component(NSCalendar.Unit.day, from: devotionDay)
```
AlexRamey: this needs to be moved inside the devotion class

iOS Client/Utilities/Formatters.swift
```diff
@@ -0,0 +1,53 @@
...
+            return "July"
+        case 8:
+            return "August"
+        case 9:
+            return "September"
+        case 10:
+            return "October"
+        case 11:
+            return "November"
+        case 12:
```
AlexRamey: default could be December

AlexRamey: i have a feeling this functionality exists for us somewhere . . .

AlexRamey: Try `DateFormatter().monthSymbols[monthInt-1]`

AlexRamey: monthSymbols returns `["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"]` for me in a playground

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
...
+            let trigger = UNCalendarNotificationTrigger(dateMatching: dateComponents, repeats: true)
+            let content = UNMutableNotificationContent()
+            content.title = devotion.verse
+            content.body = "Read your daily devotion from Mockingbird"
```
AlexRamey: we should subclass UNMutableNotificationContent and do this setup in an `init`

iOS Client/TabController/AppCoordinator.swift
```diff
@@ -43,6 +44,51 @@ class AppCoordinator: NSObject, Coordinator, UITabBarControllerDelegate, StoreSu
             self.addChildCoordinator(childCoordinator: coord)
             return coord.rootViewController
         })
+        promptForNotifications()
+    }
+    
+    func promptForNotifications() {
```
AlexRamey: should be devotion coordinator concern, see other PR comments for suggested flow

iOS Client/Store/MBStore.swift
```diff
@@ -22,6 +22,21 @@ class MBStore: NSObject {
...
+    static func loadState() -> MBAppState {
+        let defaults = UserDefaults.standard
+        var notificationPermission: Permission = .unprompted
+        if let notificationSetting = defaults.object(forKey: "notifications") as? Bool {
+            notificationPermission = notificationSetting ? .approved : .denied
+        }
+        return MBAppState(navigationState: MBNavigationState(), articleState: MBArticleState(), devotionState: MBDevotionState(), settingsState: MBSettingsState(notificationPermission: notificationPermission))
+    }
+    
+    static func saveState(state: AppState) {
```
AlexRamey: this and loadState appear to be unusued

iOS Client/Models/MBDevotion.swift
```diff
@@ -9,9 +9,14 @@
...
 
-struct MBDevotion: Codable {
+struct MBDevotion: Codable, Detailable {
     var date: String
     var text: String
     var verse: String
     var verseText: String
 }
+
+struct LoadedDevotion: Codable, Detailable {
```
AlexRamey, line 19: Hmmm, I don't think we should have a distinction between these two types

iOS Client/Store/MBStore.swift
```diff
@@ -96,35 +129,60 @@ class MBStore: NSObject {
...
-        }
-        let path = url.path.appending("/devotions")
-        let pathUrl = URL(fileURLWithPath: path)
-        if !FileManager.default.fileExists(atPath: path) {
-            client.getDevotionsWithCompletion { data, error in
-                if error == nil, let data = data {
-                    self.save(data, pathUrl) { error in
-                        if error != nil {
-                            print("Could not save devotions data to disk")
+    func markDevotionAsRead(date: String, completion: @escaping (Error?) -> Void) {
```
AlexRamey: I think this could be done in a simpler way. Why de-couple the devotions from the read dates? Whether or not a devotion is `read` is a property of that devotion. It should be part of our MBDevotion model. 

When we first download the devotions data, we shouldn't blindly save it. We should get it from the external source (whether that's the internet or a file in the app bundle), deserialize it into our Model Objects, which can have extra fields beyond what's in the JSON like `isRead` that will default to `false` and then serialize the array of our model devotions into a file at the save path. 

The only issue is if we want to make the apps reload the devotions from the external source for some reason. In that case, we would need a helper function to merge what's new with what we have, but that's better in my opinion than maintaining dates separately and dealing with two different devotion models

iOS Client/Store/MBStore.swift
```diff
@@ -96,35 +129,60 @@ class MBStore: NSObject {
...
-            completion(nil, MBDeserializationError.fetchError(msg: "Could not fetch devotions"))
         }
     }
     
+    
+    private func read<T: Codable>(fromPath: String, _ resource: T.Type, completion: @escaping (T?, Error?) -> Void) {
+        do {
+            let (manager, path, _) = try urlPackage(forPath: "/\(fromPath)")
+            if !manager.fileExists(atPath: path), let file = Bundle.main.url(forResource: fromPath, withExtension: "json") {
+                    let data = try Data(contentsOf: file)
```
AlexRamey: I miss the notion you had before of moving the data from the external source to a new location inside our document directory. I think it's important to decouple the data source from our internal store

AlexRamey (changes requested):
-----

AlexRamey:
---

iOS Client/Models/MBDevotion.swift
AlexRamey: As per discussion, I was wrong about this point. We absolutely need two different structs. One that defines the contract between us and the external data source, and another that is our internal domain model.

They should be independent  (so our domain model should not wrap the contract model but just copy over its fields) so that we can insulate our app domain from changes to the external contract. For many changes, we can just update our conversion function that converts the contract model to our domain model.

iOS Client/Utilities/Formatters.swift
jdwitten: Boom didn't know this existed nice

jdwitten:
This bad boy is ready for another review. I addressed most of your comments and reworked some of the devotions syncing. I realize it is a little bit hard to follow in MBStore but I think it will look much nicer if we get something like PromiseKit so we can handle any errors in one place

iOS Client/Store/MBStore.swift
```diff
@@ -78,11 +78,65 @@ class MBStore: NSObject {
     
 
     //New function for devotions until we figure out if this will have to go over network or can be stored locally
-    func syncDevotions(completion: @escaping ([MBDevotion]?, Error?) -> Void) {
-        loadDevotions { (devotions, devotionError) in
-            completion(devotions, devotionError)
+    func syncDevotions(completion: @escaping ([LoadedDevotion]?, Error?) -> Void) {
```
AlexRamey: I think this is still too complex. So conceptually, the `sync` part of this seems to be syncing the devotions data stored locally with the readDates file. (There is some notion of syncing with the external store, but really that's a lazy fetch of the devotions data in case we don't have it locally)

Despite a potential performance hit, I propose we do away with the extra readDatesFile and just store the array of LoadedDevotion objects. This means that marking a devotion as read is reading out the entire array of all devotions, finding the one of interest and marking it as read, then saving the array back to disk. I think that's better than maintaining two files.

I then think this method could be decomposed into two. `getDevotions` that blindly reads LoadedDevotions from documents directory and synchronously returns them, and `syncDevotions` that pulls them from the external store, converts them to LoadedDevotions, and saves them to documents directory (as well as returns them in a completion block) as you've done here. 

Then, everywhere in our app we need devotions, we can just call `getDevotions` without needing a completion block. How the sync initially gets kicked off is up for discussion. Maybe our state will let us know if they've been loaded yet. Maybe getDevotions can just return an empty list and kick off a sync job that triggers a store action on completion. Not entirely sure, but I do think it's important to

1) eliminate the extra readDatesFile and
2) create a synchronous `getDevotions`

iOS Client/Store/MBStore.swift
```diff
@@ -96,42 +150,69 @@ class MBStore: NSObject {
...
-        let path = url.path.appending("/devotions")
-        let pathUrl = URL(fileURLWithPath: path)
-        if !FileManager.default.fileExists(atPath: path) {
-            client.getDevotionsWithCompletion { data, error in
-                if error == nil, let data = data {
-                    self.save(data, pathUrl) { error in
-                        if error != nil {
-                            print("Could not save devotions data to disk")
-                        }
+    func markDevotionAsRead(date: String, completion: @escaping (Error?) -> Void) {
```
AlexRamey: why does this use a completion block? Can't it just return an error type directly since there's nothing async about it?

iOS Client/Store/MBStore.swift
```diff
@@ -96,42 +150,69 @@ class MBStore: NSObject {
...
+        let manager = FileManager.default
+        guard let url = manager.urls(for: .documentDirectory, in: .userDomainMask).first as URL? else {
+            throw StoreError.urlError(msg: "Could not create path")
+        }
+        let path = url.path.appending("/\(forPath)")
+        let pathUrl = URL(fileURLWithPath: path)
+        return (manager, path, pathUrl)
+    }
+    
     private func parse<T: Codable>(_ data: Data, _ resource: T.Type, _ completion: @escaping (T?, Error?) -> Void) {
```
AlexRamey: why completion block? sync code should just return a value directly

iOS Client/Store/MBStore.swift
```diff
@@ -96,42 +150,69 @@ class MBStore: NSObject {
...
             }
-        } else if let data = FileManager.default.contents(atPath: path) {
-            parse(data, [MBDevotion].self, completion)
-        } else {
-            completion(nil, MBDeserializationError.fetchError(msg: "Could not fetch devotions"))
         }
     }
     
+    
+    private func read<T: Codable>(fromPath: String, _ resource: T.Type, completion: @escaping (T?, Error?) -> Void) {
```
AlexRamey: why completion block? sync code  should just return a value directly

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -64,5 +75,20 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
     func numberOfSections(in tableView: UITableView) -> Int {
         return 1
     }
+    
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let devotion = devotions[indexPath.row]
+        MBStore.sharedStore.dispatch(SelectedDevotion(devotion: devotion.devotion))
+        let store = MBStore()
+        store.markDevotionAsRead(date: devotion.devotion.date) { error in
```
AlexRamey: This shouldn't have the look and feel of an async call.

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -64,5 +75,20 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
+            if error == nil {
+                store.syncDevotions { devotions, error in
+                    if error == nil, let loadedDevotions = devotions {
+                        MBStore.sharedStore.dispatch(LoadedDevotions(devotions: .loaded(data: loadedDevotions)))
```
AlexRamey: sketchy to do this in a completion block b/c you don't know what thread you're on. This needs to be in a dispatch to the main thread (or, my vote is to make markDevotionAsRead just return an error type and then you can check the error and do this imperatively on the next line. No need for this completion block approach IMO since this is not async code)

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -12,11 +12,15 @@ import ReSwift
...
-    var devotions: [MBDevotion] = []
+    var devotions: [LoadedDevotion] = []
+    var cellReusableId: String = "DevotionTableViewCell"
     override func viewDidLoad() {
         super.viewDidLoad()
         tableView.delegate = self
         tableView.dataSource = self
+        tableView.rowHeight = UITableViewAutomaticDimension
+        
+        tableView.register(UINib(nibName: cellReusableId, bundle: nil), forCellReuseIdentifier: cellReusableId)
```
AlexRamey, line 25: it's kind of a coincidence that the nibName is the same as the cellReusableId. I'd still hard-code the nibName so the reuse ID value can change without hecking up this code

AlexRamey (changes requested):

jdwitten:
Alright let's give this another shot. I think I like how this one ended up. Here are some changes I made:

1. Got rid of sycWithReadDates and just save LoadedDevotions in document file
2. Added a getDevotions method that just reads blindly from the file
3. Made a bunch of methods synchronous without completion blocks (not sure why I thought they had to be async)
4. Remove markDevotionAsRead() from the store and just added a new method to just save some LoadedDevotions to file. So the reducer now marks devotion as read when it is selected and then all devotions are saved

iOS Client/Utilities/Formatters.swift
```diff
@@ -0,0 +1,26 @@
...
+        formatter.dateFormat = "yyyy-MM-DD"
+        return formatter
+    }()
+    
+    static let calendar: NSCalendar? = {
+       return NSCalendar.init(calendarIdentifier: NSCalendar.Identifier.gregorian)
+    }()
+    
+    static func getMonth(fromInt: Int) -> String {
+        return Formatters.devotionDateFormatter.monthSymbols[fromInt-1]
```
AlexRamey: 👍

iOS Client/DevotionsController/DevotionDetailViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+        super.viewWillDisappear(animated)
+        MBStore.sharedStore.unsubscribe(self)
+    }
+
+    func newState(state: MBAppState) {
+        if let devotion = state.devotionState.selectedDevotion,
+            let date = Formatters.devotionDateFormatter.date(from: devotion.date),
+            let monthInt = Formatters.calendar?.component(.month, from: date),
+            let day = Formatters.calendar?.component(.day, from: date) {
+            monthLabel.text = Formatters.getMonth(fromInt: monthInt)
```
AlexRamey: Can be done in a later refactor, but I think It'd be cool to add monthName and monthDay (idk if those are the best names, maybe like displayDateMonthComponent displayDateDayComponent . . .) as lazily computed string properties on LoadedDevotion that encapsulate all this logic.

iOS Client/DevotionsController/DevotionNotification.swift
```diff
@@ -0,0 +1,26 @@
...
+    
+    init(devotion: MBDevotion) {
+        self.devotion = devotion
+        super.init()
+        self.title = devotion.verse
+        self.body = "Read your daily devotion from Mockingbird"
+        self.sound = UNNotificationSound.default()
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
```
AlexRamey, line 23: Hmm, wonder what the deal with this is? It breaks if we delete it?

AlexRamey: the swift docs are kind of unclear. They say all subclasses have to provide an implementation or required initializers but also that `You do not have to provide an explicit implementation of a required initializer if you can satisfy the requirement with an inherited initializer.` I don't know what that means . . .

iOS Client/DevotionsController/DevotionTableViewCell.swift
```diff
@@ -0,0 +1,28 @@
...
+
+class DevotionTableViewCell: UITableViewCell {
+
+    @IBOutlet weak var dayLabel: UILabel!
+    @IBOutlet weak var monthLabel: UILabel!
+    @IBOutlet weak var verseLabel: UILabel!
+    @IBOutlet weak var readNotifier: UIImageView!
+    
+    
+    func configure(day: String, month: String, verse: String, read: Bool) {
```
AlexRamey: a minor note, but this could just accept one parameter, the LoadedDevotion

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+            store.syncDevotions { syncedDevotions, error in
+                if error != nil {
+                    MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.error))
+                } else if let newDevotions = syncedDevotions {
+                    MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: newDevotions)))
+                }
+            }
+        } else {
+            MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: devotions)))
+        }
```
AlexRamey: good stuff.

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
         let devotionsController = MBDevotionsViewController.instantiateFromStoryboard()
         navigationController.pushViewController(devotionsController, animated: true)
         MBStore.sharedStore.subscribe(self)
+        let store = MBStore()
+        let devotions = store.getDevotions()
+        if devotions.count == 0 {
+            store.syncDevotions { syncedDevotions, error in
+                if error != nil {
+                    MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.error))
```
AlexRamey: To be safe, these two dispatches should be in a dispatch.main.async block to guarantee that they occur on the main thread, since that is a requirement of ReSwift. In this case, I think it would be on the main thread b/c our client method to load the devotions is probably synchronous, but if we changed that out to be a network call then we might get some random undefined behavior stemming from these dispatches occurring on a background thread.

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+                } else if let newDevotions = syncedDevotions {
+                    MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: newDevotions)))
+                }
+            }
+        } else {
+            MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: devotions)))
```
AlexRamey: this one is fine, def on the main thread

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -43,4 +84,19 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
         build(newRoute: newRoute)
         route = newRoute
     }
+    
+    // MARK: - UNNotificationDelegate
+    func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        MBStore().syncDevotions { devotions, error in
```
AlexRamey: this can just be a synchronous getDevotions

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+                } else if let newDevotions = syncedDevotions {
+                    MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: newDevotions)))
+                }
+            }
+        } else {
+            MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: devotions)))
+        }
+    }
+    
+    func promptForNotifications() {
```
AlexRamey: where does this get called?

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+        let center = UNUserNotificationCenter.current()
         
-        MBStore().syncDevotions { devotions, error in
-            if error != nil {
-                MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.error))
-            } else if let devotions = devotions {
-                MBStore.sharedStore.dispatch(LoadedDevotions(devotions: Loaded.loaded(data: devotions)))
+        center.requestAuthorization(options: [.alert, .badge, .sound]) { granted, error in
+            if granted {
+                MBStore().syncDevotions { devotions, error in
```
AlexRamey: this can be a `get`

iOS Client/DevotionsController/DevotionsCoordinator.swift
AlexRamey: maybe it could be called from a newState method when the devotions get loaded, that way we know we can use `getDevotions` in the body of prompt and schedule. Also might be worth throwing a bool in UserDefaults called `areNotificationsScheduled` or something so we can do nothing if it's already taken care of

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -12,11 +12,17 @@ import ReSwift
 class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewDelegate, UITableViewDataSource {
 
     @IBOutlet weak var tableView: UITableView!
-    var devotions: [MBDevotion] = []
+    var store: MBStore!
+    var devotions: [LoadedDevotion] = []
+    var cellReusableId: String = "DevotionTableViewCell"
```
AlexRamey, line 17: i've usually seen `reuseIdentifier`, but I'm not gonna split hairs over it (and the value would be something more like `DevotionCellReuseIdentifier`)

iOS Client/DevotionsController/MBDevotionsViewController.swift
AlexRamey: `nibName` should be hard-coded in my opinion. The nibName and the cellReuseIdentifier should be independent of one another

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -52,8 +58,15 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
-        let cell = tableView.dequeueReusableCell(withIdentifier: "DevotionTableViewCell") ?? UITableViewCell()
-        cell.textLabel?.text = devotion.text
+        //swiftlint:disable force_cast
+        let cell = tableView.dequeueReusableCell(withIdentifier: cellReusableId) as! DevotionTableViewCell
+        //swiftlint:enable force_cast
+        if let date = Formatters.devotionDateFormatter.date(from: devotion.devotion.date),
+            let day = Formatters.calendar?.components(.day, from: date).day,
+            let monthInt = Formatters.calendar?.component(.month, from: date) {
+            cell.configure(day: String(describing: day), month: Formatters.getMonth(fromInt: monthInt), verse: devotion.devotion.verse, read: devotion.read)
+
```
AlexRamey: whitespace. Could be cool to just say cell.configure(devotion) and let it handle everything. If it encounters any problems with dates, it can use empty fields. As it stands now, if a date error were to occur, the cell would have old values still in it from its previous position in the table view

AlexRamey: i've seen some quotes before in a book called Clean Code that a function should accept at most two parameters (kind of ridiculous, but something to strive for nonetheless 🐘 )

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -64,5 +77,17 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
     func numberOfSections(in tableView: UITableView) -> Int {
         return 1
     }
+    
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+       let devotion = devotions[indexPath.row]
+        MBStore.sharedStore.dispatch(SelectedDevotion(devotion: devotion))
+        do {
+            try store.saveDevotions(devotions: devotions)
```
AlexRamey: before this, where does the read property on the devotion actually get toggled to 'true' ?

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -64,5 +77,17 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
+        } catch {
+            //There was an error saving devotion as read so reverse
```
AlexRamey: nice handling of this case.

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -108,9 +108,9 @@ class MBClient: NSObject {
         }.resume()
     }
     
-    func getDevotionsWithCompletion(completion: @escaping (Data?, Error?) -> Void ) {
+    func getJSONFile(name: String, completion: @escaping (Data?, Error?) -> Void ) {
         do {
-            if let file = Bundle.main.url(forResource: "devotions", withExtension: "json") {
+            if let file = Bundle.main.url(forResource: name, withExtension: "json") {
```
AlexRamey, line 113: nice

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -108,9 +108,9 @@ class MBClient: NSObject {
         }.resume()
     }
     
-    func getDevotionsWithCompletion(completion: @escaping (Data?, Error?) -> Void ) {
+    func getJSONFile(name: String, completion: @escaping (Data?, Error?) -> Void ) {
```
AlexRamey, line 111: so, this is also synchronous, but I'm OK with it exposing an async interface since we could conceivably change it to happen in the background. This hints to callers of this code that their completion block may not run on the main thread

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+                    if error != nil {
+                        print("Could not sync devotions and schedule devotions")
+                        print(error)
+                    } else if devotions != nil {
+                        center.delegate = self
+                        self.scheduleNotifications(devotions: devotions!.map{$0.devotion})
```
AlexRamey: this code seems out-dated, since the LoadedDevotions no longer contain an embedded MBDevotion

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -25,13 +26,53 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+                    }
+                }
+            } else {
+                print(error)
+            }
+        }
+    }
+    
+    func scheduleNotifications(devotions: [MBDevotion]) {
```
AlexRamey: should accept an array of LoadedDevotions

iOS Client/Models/MBDevotion.swift
```diff
@@ -9,9 +9,46 @@
...
 
 
-struct MBDevotion: Codable {
+struct MBDevotion: Codable, Detailable {
     var date: String
     var text: String
     var verse: String
     var verseText: String
+    
+    var dateComponentsForNotification: DateComponents? {
```
AlexRamey: this behavior will need to be moved down into the LoadedDevotion class

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -59,8 +65,27 @@ func devotionReducer(action: Action, state: DevotionState?) -> DevotionState {
...
+        if case Loaded.loaded(data: let devotions) = nextState.devotions {
+            let newDevotions: [LoadedDevotion] = devotions.map { devotion in
+                if devotion.date == nextState.selectedDevotion?.date {
+                    return LoadedDevotion(devotion: devotion.devotion, read: true)
+                } else {
+                    return devotion
+                }
+            }
+            nextState.devotions = Loaded.loaded(data: newDevotions)
+        }
```
AlexRamey: since we are dealing with classes (ref types), I think it would work to delete lines 70-79 and just write `nextState.selectedDevotion.read = true`

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -59,8 +65,27 @@ func devotionReducer(action: Action, state: DevotionState?) -> DevotionState {
...
+    case let action as DevotionNotification:
+        nextState.selectedDevotion = action.devotion
```
AlexRamey: maybe we should add `nextState.selectedDevotion.read = true` here as well ?

iOS Client/Store/MBStore.swift
```diff
@@ -38,6 +38,14 @@ class MBStore: NSObject {
         return performFetch(managedContext: persistentContainer.viewContext, fetchRequest: fetchRequest) as? [MBArticle] ?? []
     }
     
+    func getDevotions() -> [LoadedDevotion] {
+        do {
+            return try read(fromPath: "devotions", [LoadedDevotion].self)
+        } catch {
+            return []
+        }
+    }
```
AlexRamey, line 47: very simple. I like it.

iOS Client/Store/MBStore.swift
```diff
@@ -78,12 +86,26 @@ class MBStore: NSObject {
...
 
     //New function for devotions until we figure out if this will have to go over network or can be stored locally
-    func syncDevotions(completion: @escaping ([MBDevotion]?, Error?) -> Void) {
-        loadDevotions { (devotions, devotionError) in
-            completion(devotions, devotionError)
+    func syncDevotions(completion: @escaping ([LoadedDevotion]?, Error?) -> Void) {
+        self.client.getJSONFile(name: "devotions") { data, error in
+            if error == nil, let data = data {
+                do {
+                    //We got some data now parse
```
AlexRamey: hells yeah we do

AlexRamey: could use a whitespace between second slash and We. this should be a lint warning if it's not

iOS Client/Store/MBStore.swift
```diff
@@ -78,12 +86,26 @@ class MBStore: NSObject {
...
+                    let loadedDevotions = devotions.map {LoadedDevotion(devotion: $0, read: false)}
+                    try self.save(loadedDevotions, "devotions")
+                    completion(loadedDevotions, nil)
+                } catch let error {
+                    completion(nil, error)
+                }
+            } else {
+                //Failed so complete with no devotions
+                completion(nil, error)
+            }
```
AlexRamey: very simple. I like it

iOS Client/Store/MBStore.swift
AlexRamey: the read helper function really shines here

iOS Client/Store/MBStore.swift
AlexRamey: the parse helper function really shines here

iOS Client/Store/MBStore.swift
```diff
@@ -96,51 +118,47 @@ class MBStore: NSObject {
         performDownload(clientFunction: client.getArticlesWithCompletion, persistentContainer: persistentContainer, deserializeFunc: MBArticle.deserialize, completion: completion)
     }
     
-    private func loadDevotions(completion: @escaping ([MBDevotion]?, Error?) -> Void) {
-        
+    func saveDevotions(devotions: [LoadedDevotion]) throws {
+        try self.save(devotions, "devotions")
+    }
+    
+    private func read<T: Codable>(fromPath: String, _ resource: T.Type) throws -> T {
```
AlexRamey: super minor, but if we want to get fancy I think we could name this parameter like `fromPath path:String` (can't remember the exact syntax) so callers get intuitive `fromPath` view of the param name but then internally this function can refer to it as just `path`, which makes more sense when calling urlPackage(forPath: path)

iOS Client/Store/MBStore.swift
```diff
@@ -96,51 +118,47 @@ class MBStore: NSObject {
...
+        let (manager, path, _) = try urlPackage(forPath: "\(fromPath)")
```
AlexRamey: if fromPath is already a string, then why do string interpolation?

iOS Client/Store/MBStore.swift
```diff
@@ -96,51 +118,47 @@ class MBStore: NSObject {
...
+        if let data = manager.contents(atPath: path) {
+            return try parse(data, T.self)
+        } else {
+            throw StoreError.readError(msg: "No data could be read from file \(fromPath)")
+        }
+    }
+    
+    private func urlPackage(forPath: String) throws -> (FileManager, String, URL) {
```
AlexRamey: same thing here, could expose forPath to callers and name the parameter just path for internal usuage

iOS Client/Store/MBStore.swift
```diff
@@ -96,51 +118,47 @@ class MBStore: NSObject {
...
     
-    private func save(_ data: Data, _ pathUrl: URL, _ completion: @escaping (Error?) -> Void) {
+    private func save<T: Encodable>(_ data: T, _ path: String) throws {
         do {
-            try data.write(to: pathUrl, options: [.atomic])
-            return completion(nil)
+            let (_, _, pathUrl) = try self.urlPackage(forPath: path)
+            let encoder = JSONEncoder()
+            let encodedData = try encoder.encode(data)
+            try encodedData.write(to: pathUrl, options: [.atomic])
```
AlexRamey, line 196: nice usage of `.atomic`

iOS Client/Store/MBStore.swift
```diff
@@ -233,3 +251,10 @@ class MBStore: NSObject {
         }
     }
 }
+
+enum StoreError: Error {
+    case readError(msg: String)
+    case writeError(msg: String)
+    case parseError(msg: String)
+    case urlError(msg: String)
+}
```
AlexRamey, line 297: nice

AlexRamey (changes requested):
Some very nice improvements. I think this is starting to take the shape we want. 

I went through it very thoroughly this time, now that the structure is more in place. 36 comments, but most should be pretty quick to address.

iOS Client/DevotionsController/DevotionNotification.swift
jdwitten: Yeah I'm a little confused about this. I don't think this ever actually would get called but it would be nice to not include it

iOS Client/DevotionsController/DevotionTableViewCell.swift
jdwitten: nice I like that

iOS Client/DevotionsController/DevotionsCoordinator.swift
jdwitten: good call I should put that first dispatch in an async block

iOS Client/DevotionsController/DevotionsCoordinator.swift
jdwitten: whoops this used to get called in the DevotionsCoordinator start method but must have gotten lost in the shuffle. Yeah I like having a field in UserDefaults I can add that

iOS Client/DevotionsController/DevotionsCoordinator.swift
jdwitten: good call

iOS Client/DevotionsController/DevotionsCoordinator.swift
jdwitten: yep forgot about that one

iOS Client/DevotionsController/MBDevotionsViewController.swift
jdwitten: So that happens in the reducer during the SelectedDevotion action

iOS Client/HTTPClient/MBClient.swift
jdwitten: Yeah I'm going to keep it like this just in case in the future we have to get this file over the network

iOS Client/Reducers/MBAppReducer.swift
jdwitten: pretty sure these are all structs though. DevotionState is a struct and devotions are structs

iOS Client/Reducers/MBAppReducer.swift
jdwitten: Yes this is a good catch definitely need it here

iOS Client/Store/MBStore.swift
jdwitten: Ah just left over from earlier we don't need to do string interpolation any more

iOS Client/Reducers/MBAppReducer.swift
jdwitten: Made this more succinct in the latest commits

iOS Client/DevotionsController/DevotionsCoordinator.swift
jdwitten: I'm going to keep the promptForNotifications call after we sync devotions on the first start up of the devotions coordinator. It's not going to do anything if we have already scheduled or if the user has already been prompted. Not sure exactly where it would go in newState because we wouldn't want to schedule every time we got a new state. Could add a new property on DevotionState that said whether or not we have scheduled yet. I'll leave it for now and we can refactor if necessary.

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -43,4 +84,14 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber {
         build(newRoute: newRoute)
         route = newRoute
     }
+    
+    // MARK: - UNNotificationDelegate
+    func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        let devotions = MBStore().getDevotions()
+        let date = Date()
+        if let devotion = devotions.first(where: { Formatters.devotionDateFormatter.date(from: $0.date) == date}) {
```
AlexRamey: This is a little dangerous. You're betting that the current date (line 91) exactly matches the devotion's date. Maybe it would be better to convert the current `date` to a string in YYYY-MM-DD format and then do string comparison

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -59,8 +65,27 @@ func devotionReducer(action: Action, state: DevotionState?) -> DevotionState {
...
+            let index = devotions.index(where: { $0.date == action.devotion.date }) {
+            devotions[index].read = true
+            nextState.devotions = Loaded.loaded(data: devotions)
+        }
+    case let action as DevotionNotification:
+        nextState.selectedDevotion = action.devotion
+        if case Loaded.loaded(data: var devotions) = nextState.devotions,
+            let index = devotions.index(where: { $0.date == action.devotion.date }) {
+            devotions[index].read = true
+            nextState.devotions = Loaded.loaded(data: devotions)
```
AlexRamey: `lines 69-73` and `76-80` are identical. We need to factor this out into a helper of some sort

iOS Client/DevotionsController/MBDevotionsViewController.swift
AlexRamey: Hmm, I am OK with this for now, but I think it needs to be clearer in an upcoming refactor when we mark a devotion as read

The concerning thing is that the call to dispatch has a side effect of updating this view controller's devotions property that the next lines rely on.

I think it would be clearer to only do the dispatch here, then in `newState` check if the selected devotion in the devotions property is currently marked as false and, if so, call a helper to do the save

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -64,5 +72,17 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
+            print("Error marking devotion as read")
+            devotions[indexPath.row].read = false
+            MBStore.sharedStore.dispatch(LoadedDevotions(devotions: .loaded(data: devotions)))
```
AlexRamey: Works for now; in a multi-threaded setting we'd want to be careful about clobbering all the devotions in the state with this view controller's local copy, b/c something could change in between the dispatch selected devotion action and this . . . I guess the solution would be an additive only approach where we dispatch a new action called UnselectDevotion or something

iOS Client/Models/MBDevotion.swift
```diff
@@ -9,9 +9,64 @@
...
+    
+    init(devotion: MBDevotion, read: Bool) {
+        date = devotion.date
+        text = devotion.text
+        verse = devotion.verse
+        verseText = devotion.verseText
+        self.read = read
+    }
+    
+    var devotion: MBDevotion {
```
AlexRamey: where is this used? We should never really go back once we have our domain model.

iOS Client/DevotionsController/DevotionsCoordinator.swift
AlexRamey: a moot point, but you could use the new lazy `day` property on devotion instead of doing the Formatters call here . . .

AlexRamey: `$0.day == date` would be equivalent; but, again, I think we should do string comparison of YYYY-MM-DD instead of date comparison

iOS Client/Reducers/MBAppReducer.swift
AlexRamey: Or, equivalently, why do we need the DevotionNotification action? Couldn't we just re-use SelectedDevotion and long as you add `nextState.selectedTab = .devotions` to the navigation reducer for SelectedDevotion action?

iOS Client/Utilities/Formatters.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Formatters.swift
+//  iOS Client
+//
+//  Created by Jonathan Witten on 11/14/17.
+//  Copyright © 2017 Mockingbird. All rights reserved.
+//
+
+import Foundation
+
+struct Formatters {
```
AlexRamey: double-check me, but now that we're encapsulating a lot of this work in the LoadDevotion struct, couldn't we remove this file all-together and do these  as private things inside of LoadedDevotion struct? 

(Looks like calendar was already moved over)

1 good use of this file might be a helper that goes from Date to String in YYYY-MM-DD format since that's what we may use for comparison in the notification delegate.

iOS Client/DevotionsController/DevotionsCoordinator.swift
AlexRamey: what you did is good

iOS Client/DevotionsController/DevotionDetailViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        MBStore.sharedStore.unsubscribe(self)
+    }
+
+    func newState(state: MBAppState) {
+        if let devotion = state.devotionState.selectedDevotion,
+            let date = Formatters.devotionDateFormatter.date(from: devotion.date),
+            let monthInt = Formatters.calendar?.component(.month, from: date),
```
AlexRamey: there are now helpers for this stuff in LoadedDevotion struct you can use here

AlexRamey (changes requested):

iOS Client/DevotionsController/DevotionsCoordinator.swift
jdwitten: Yeah I think you're right we should compare the date strings not the actually date objects

iOS Client/DevotionsController/MBDevotionsViewController.swift
jdwitten: I agree that it isn't ideal that the view controllers devotions rely on that dispatched action. Although, I don't think saving in the reducer is a good idea. Reducers shouldn't have any side effects like saving data or fetching data, they should just take a state and an action and return a new state, nothing else. I'll try to think of a better solution but might be good to just get this in.

iOS Client/Models/MBDevotion.swift
jdwitten: I don't think I need this I can remove

iOS Client/Reducers/MBAppReducer.swift
jdwitten: I like the second option I can do that

iOS Client/Utilities/Formatters.swift
jdwitten: I think in the future if we needed some formatters that were shared between a couple screens or models than we could use this file. I'm ok with having all of this in LoadedDevotion for now

iOS Client/State/MBState.swift
```diff
@@ -46,11 +46,19 @@ struct MBArticleState: ArticleState {
 protocol DevotionState {
     var devotions: Loaded<[LoadedDevotion]> { get set }
     var selectedDevotion: LoadedDevotion? { get set }
+    mutating func mark(devotion: LoadedDevotion, asRead read: Bool)
```
AlexRamey: nice use of mutating to get ref type semantics

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -59,8 +63,18 @@ func devotionReducer(action: Action, state: DevotionState?) -> DevotionState {
     switch action {
     case let action as LoadedDevotions:
         nextState.devotions = action.devotions
+    case let action as SelectedDevotion:
+        nextState.selectedDevotion = action.devotion
+        nextState.mark(devotion: action.devotion, asRead: true)
+    case let action as UnreadDevotion:
+        nextState.mark(devotion: action.devotion, asRead: false)
```
AlexRamey, line 70: really good man

iOS Client/Store/MBStore.swift
```diff
@@ -96,51 +118,55 @@ class MBStore: NSObject {
...
+    func saveDevotions(devotions: [LoadedDevotion]) throws {
+        try self.save(devotions, forPath: "devotions")
+    }
+    
+    func replace(devotion: LoadedDevotion) throws {
+        let devotions = try self.read(fromPath: "devotions", [LoadedDevotion].self)
+        let markedDevotions = devotions.map { oldDevotion in
+            oldDevotion.date == devotion.date ? devotion : oldDevotion
+        }
+        try self.save(markedDevotions, forPath: "devotions")
```
AlexRamey, line 159: nice

AlexRamey (dismissed):
Very nice. This code looks good to me. Look forward to playing with it. I like the idea about doing a middleware to persist the devotions (and maybe to initially read them?) but this is good enough for our purposes. Good work. Let's get this merged and I can hopefully make so contributions to the articles screen this week. DZ wants to get lunch Friday ;)

AlexRamey (approved):
merged

PR #22, opened 2017-12-06
1. Articles table view shows one cell per category; the cell contains…

AlexRamey:
… a collection view of horizontally scrolling articles from that category


iOS Client/ArticlesController/TableViewCell/ArticleTableViewCell.swift
```diff
@@ -0,0 +1,77 @@
...
+        self.collectionView.reloadData()
+    }
+    
+    func setupLayout(forItemCount itemCount: Int) {
+        guard let layout = self.collectionView.collectionViewLayout as? UICollectionViewFlowLayout else {
+            return
+        }
+        
+        let screenWidth = UIScreen.main.bounds.width
+        if itemCount == 1 {
```
jdwitten, line 48: nice, I like it make the article big if there is only one

iOS Client/ArticlesController/TableViewCell/ArticleTableViewCell.swift
```diff
@@ -0,0 +1,77 @@
...
+        
+        let layout = UICollectionViewFlowLayout()
+        layout.scrollDirection = .horizontal
+        layout.minimumLineSpacing = 8.0 // space between columns
+        layout.minimumInteritemSpacing = 500.0 // space between rows (big to keep it all on one row)
+        layout.sectionInset = UIEdgeInsets(top: 0.0, left: 8.0, bottom: 0.0, right: 8.0)
+        self.collectionView.collectionViewLayout = layout
+    }
+    
+    func configure(articles: [MBArticle]) {
```
jdwitten, line 35: I like the collection view inside of the table view cell. Only thing I'm thinking from a UX experience it might be difficult to see only the most recent articles. Maybe have a section that is just new articles? No actionable suggestions just something we should think about

iOS Client/ArticlesController/CollectionViewCell/ArticleItem.swift
```diff
@@ -0,0 +1,38 @@
...
+    @IBOutlet weak var coverImage: UIImageView!
+    @IBOutlet weak var titleLabel: UILabel!
+    let client = MBClient()
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        // Initialization code
+    }
+    
+    func configure(article: MBArticle) {
```
jdwitten, line 22: could be cool to throw in some rounded corners or something snazzy on these images. Don't have to do it for this PR

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -120,7 +115,7 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
                 }))
             
                 for topLevelCategory in topLevelCategories {
```
jdwitten: Do we need to do this second for loop here? I'm thinking we are already looping through categories in the flatMap above, so instead of creating that Set and instead of returning, just insert the category into retVal and update its corresponding array. Let me know if that makes sense or if I'm missing something

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -134,6 +129,22 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
                 print("Excluding article \(article.title ?? "<no title>") b/c it has no categories")
             }
         }
+        
+        // sort newest articles first
```
jdwitten: Learned about something cool today called Date.distantPast that I think we could use here to make this loop a little more succinct maybe something like:
`
retVal.keys.forEach {
    retVal[$0] = retVal[$0]?.sorted { a1, a2 in
        return (a1 .date as? Date ?? Date.distantPast) > (a2.date as? Date ?? Date.distantPast)
    }
}
`

jdwitten: whoops the formatting turned out bad on that ^

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -15,7 +15,7 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
     var articlesByCategory = [String: [MBArticle]]()
     var topLevelCategories: [String] = []
     let client = MBClient()
-    static let ArticleTableViewCellId = "ArticleTableViewCell"
+    let reuseIdentifier = "ArticleCellReuseIdentifier"
```
jdwitten, line 18: nice

jdwitten (changes requested):
looks good, just a few comments

iOS Client/ArticlesController/CollectionViewCell/ArticleItem.swift
AlexRamey: agreed, we need to make this as fancy as possible. sounds like a good follow-up.

iOS Client/ArticlesController/TableViewCell/ArticleTableViewCell.swift
AlexRamey: yeah, the website somehow has featured articles but I don't see that exposed to us in the API. Would be great to get a featured section or New Content section or something. Let's discuss this on Friday. For now, I am sorting the newest articles first so the user should at least see them for each section

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: I like that; this felt very clunky to me when I was writing it.

jdwitten (dismissed):
Don't know if you saw my other comment about having two loops when building out the categories, but I think it's something to look at for a future refactor. I'll approve now but let's revisit

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: addressed.

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: I cleaned this up a bit. I couldn't see a way to get rid of the two loops b/c I needed to create a set before continuing to remove potential duplicates.

I moved these details into a helper on article called `getTopLevelCategories` that just returns a set. So we just iterate over that set directly here now

jdwitten (approved):
Looks good my dude much better

PR #23, opened 2017-12-07
adds a bookmark flag to MBArticle and a method to set this field in t…

jdwitten:
* Adds a bookmark flag to MBArticle and a method to set this field in the Store
* Also reuses MBArticleViewController for bookmarks

(read my TODO comment on the bookmark method)

### DIFF
--- a/iOS Client/ArticlesController/MBArticleDetailViewController.swift
+++ b/iOS Client/ArticlesController/MBArticleDetailViewController.swift
@@ -27,6 +27,12 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         super.viewDidLoad()
         configureWebView()
         configureBackButton()
+        configureBookmarkButton()
+    }
+    
+    func configureBookmarkButton() {
+        let barButton = UIBarButtonItem(title: "Bookmark", style: .plain, target: self, action: #selector(MBArticleDetailViewController.bookmark(_:)))
+        self.navigationItem.rightBarButtonItem = barButton
     }
     
     func configureWebView() {
@@ -34,7 +40,6 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         webView = WKWebView(frame: .zero, configuration: webConfiguration)
         webView.navigationDelegate = self
         view = webView
-        
         if let content = self.selectedArticle?.content {
             print(content)
             let cssHead: String = "<head>\n<link rel=\"stylesheet\" type=\"text/css\" href=\"MB.css\">\n</head>\n"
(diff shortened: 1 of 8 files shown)


iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -53,6 +58,23 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         }
     }
     
+   @objc func bookmark(_ sender: Any) {
```
AlexRamey, line 61: is this @objc stuff necessary. Can't we do an IBAction? Idk about this, it's been a while

iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -53,6 +58,23 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
...
+        do {
+            guard let appDelegate = UIApplication.shared.delegate as? AppDelegate,
+                let container = appDelegate.persistentContainer,
+                let article = selectedArticle else {
+                return
+            }
+            
+            try MBStore().bookmark(article: article, persistentContainer: container)
+            // TODO: Dispatch an action here to update state if we change our models to be structs
```
AlexRamey, line 70: good point, I'll make this my highest priority after the image loading

iOS Client/BookmarksController/BookmarksCoordinator.swift
```diff
@@ -25,7 +25,7 @@ class BookmarksCoordinator: NSObject, Coordinator, StoreSubscriber {
     
     // MARK: - Coordinator
     func start() {
-        let bookmarksController = MBBookmarksViewController.instantiateFromStoryboard()
+        let bookmarksController = MBArticlesViewController.instantiateFromStoryboard(bookmarks: true)
```
AlexRamey, line 28: I don't know about re-using the same VC man. Let's discuss this tomorrow.

iOS Client/Store/MBStore.swift
```diff
@@ -38,6 +38,19 @@ class MBStore: NSObject {
...
     
+    func getBookmarkedArticles(persistentContainer: NSPersistentContainer) -> [MBArticle]{
+        let fetchRequest = NSFetchRequest<NSManagedObject>(entityName: MBArticle.entityName)
+        let predicate = NSPredicate(format: "bookmarked == TRUE")
+        fetchRequest.predicate = predicate
+        return performFetch(managedContext: persistentContainer.viewContext, fetchRequest: fetchRequest) as? [MBArticle] ?? []
+
+    }
+    
+    func bookmark(article: MBArticle, persistentContainer: NSPersistentContainer) throws {
```
AlexRamey, line 49: hmmm, this is assuming the MBArticle NSManagedObject is linked to the Persistent Container's View context. This happens to be the case, but it's a little dangerous to assume so.

I think you can just pass the article here, and use its managedObjectContext property to get the context, on which you can call save().

AlexRamey (commented):
Left some comments for now, let's talk with DZ tomorrow about how this should look. I'm a little unsure about re-using the VC we have for articles home.

jdwitten:
Closing this because we will implement bookmark with its own VC

PR #24, opened 2017-12-08
schedules a notification to appear tomorrow at lunch

jdwitten:


iOS Client/Models/MBDevotion.swift
```diff
@@ -58,7 +58,7 @@ struct LoadedDevotion: Codable, Detailable {
             return nil
         }
         var dateComponents = DateComponents()
-        dateComponents.hour = 10
+        dateComponents.hour = 12
```
AlexRamey, line 61: haha let's go! cool man

iOS Client/Models/MBDevotion.swift
```diff
@@ -72,7 +72,7 @@ struct LoadedDevotion: Codable, Detailable {
     
     static let devotionDateFormatter: DateFormatter = {
         let formatter = DateFormatter()
-        formatter.dateFormat = "yyyy-MM-DD"
+        formatter.dateFormat = "yyyy-MM-dd"
```
AlexRamey, line 75: nice, something did feel fishy about this. Dates are always a huge heck

iOS Client/Models/MBDevotion.swift
AlexRamey: I guess all the stuff about time zones will "just work" Idk. we'll see. Maybe some localized logic is somewhere

AlexRamey (approved):

PR #25, opened 2017-12-09
Feature/image saving

AlexRamey:
this is a pretty tiny PR. I also got started on some refactoring though so the diff shows the Copy-Paste of all our Custom Model files into the Core Data Class files (only the Properties ones get overwritten when you do code generation)


iOS Client/ArticlesController/CollectionViewCell/ArticleItem.swift
```diff
@@ -13,26 +13,43 @@ class ArticleItem: UICollectionViewCell {
...
-        
+        self.articleID = article.articleID
         self.coverImage.image = nil
+        
         titleLabel.attributedText = article.title?.convertHtml()
         titleLabel.font = UIFont.systemFont(ofSize: 18)
         titleLabel.textColor = UIColor.ArticleTitle
         titleLabel.sizeToFit()
+        
+        if let savedData = article.image?.image {
```
jdwitten, line 32: nice looks good

jdwitten (approved):

PR #28, opened 2017-12-10
Refactor- Break up MBStore into smaller components

AlexRamey:
1. Broke our huge MBStore file into 3 files: MBStore, MBArticlesStore, and MBDevotionsStore 

2. Removed the year date component from devotion.dateComponentsForNotification so they will work indefinitely


iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -15,6 +15,7 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
     var articlesByCategory = [String: [MBArticle]]()
     var topLevelCategories: [String] = []
     let client = MBClient()
+    let articlesStore = MBArticlesStore()
```
jdwitten, line 18: This is a good call so we don't have to instantiate new stores everywhere. One thing we do on Synchrony is inject these dependencies via a "build" method which would be our version of instantiateFromStoryboard() since we don't have inits for these view controllers. Might be something to consider for a future refactor

iOS Client/Store/MBDevotionsStore.swift
```diff
@@ -0,0 +1,107 @@
...
+    
+    func replace(devotion: LoadedDevotion) throws {
+        let devotions = try self.read(fromPath: "devotions", [LoadedDevotion].self)
+        let markedDevotions = devotions.map { oldDevotion in
+            oldDevotion.date == devotion.date ? devotion : oldDevotion
+        }
+        try self.save(markedDevotions, forPath: "devotions")
+    }
+    
+    private func read<T: Codable>(fromPath path: String, _ resource: T.Type) throws -> T {
```
jdwitten, line 68: I might break these out into another shared file when I do the Podcasts stuff

jdwitten (approved):
Looks good my dude

PR #30, opened 2017-12-11
1. Rolled own Core Data Stack 2. Pass managed context from app delegate

AlexRamey:
This still needs manual testing to verify nothing is broken with normal app usage and also background app refresh.

It seems that mockingbird's https is down (SSL cert issue) at the moment, and when I hit the regular http: endpoint, the Wordpress api isn't returning the paging headers . . . so I guess we hope it's fixed soon?

### DIFF
--- a/iOS Client/CoreDataStack/CoreDataStack.swift
+++ b/iOS Client/CoreDataStack/CoreDataStack.swift
@@ -1,0 +1,45 @@
+//
+//  CoreDataStack.swift
+//  iOS Client
+//
+//  Created by Alex Ramey on 12/10/17.
+//  Copyright © 2017 Mockingbird. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+class CoreDataStack {
+    private let modelName: String
+    
+    init(modelName: String) {
+        self.modelName = modelName
+    }
+    
+    private lazy var storeContainer: NSPersistentContainer = {
+        let container = NSPersistentContainer(name: self.modelName)
+        container.loadPersistentStores(completionHandler: { (storeDescription, error) in
+            if let error = error as NSError? {
+                print("Unresolved error \(error), \(error.userInfo)")
+            }
+        })
+        return container
+    }()
+    
+    lazy var managedContext: NSManagedObjectContext = {
+        return self.storeContainer.viewContext
+    }()
+    
+    lazy var privateQueueContext: NSManagedObjectContext = {
+       return self.storeContainer.newBackgroundContext()
+    }()
+    
+    func saveContext() {
+        guard managedContext.hasChanges else { return }
+        do {
+            try managedContext.save()
+        } catch let error as NSError {
+            print("Unresolved error \(error), \(error.userInfo)")
+        }
+    }
+}
--- a/iOS Client/Store/MBArticlesStore.swift
+++ b/iOS Client/Store/MBArticlesStore.swift
@@ -131,6 +131,7 @@ class MBArticlesStore: NSObject {
             }
             
             var isNewData = false
+            var caughtError: Error? = nil
             // data is an array of Data, where each datum is a serialized array representing a page of results
             // thus, think of data as an array of results pages
             for jsonData in data {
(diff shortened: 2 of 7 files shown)


iOS Client/CoreDataStack/CoreDataStack.swift
```diff
@@ -0,0 +1,45 @@
...
+//  iOS Client
+//
+//  Created by Alex Ramey on 12/10/17.
+//  Copyright © 2017 Mockingbird. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+class CoreDataStack {
```
jdwitten, line 12: nice I like getting all of this out of AppDelegate

iOS Client/Store/MBArticlesStore.swift
```diff
@@ -17,184 +17,172 @@ class MBArticlesStore: NSObject {
...
+                print("Could not fetch. \(error), \(error.userInfo)")
+                retVal = []
+            }
         }
+        
+        return retVal
     }
     
-    /***** Download Data from Network (MAY BE INVOKED FROM ANY THREAD) *****/
-    func syncAllData(persistentContainer: NSPersistentContainer, completion: @escaping (Bool?, Error?) -> Void) {
```
jdwitten, line 46: I think if we introduce PromiseKit or PinkyPromise then this method will be much more manageable. Right now it is a little too much nesting for me

iOS Client/Store/MBArticlesStore.swift
```diff
@@ -17,184 +17,172 @@ class MBArticlesStore: NSObject {
...
                 })
             }
         }
     }
     
-    private func linkCategoriesTogether(persistentContainer: NSPersistentContainer, completion: @escaping (Error?) -> Void) {
-        persistentContainer.performBackgroundTask { (managedContext) in
-            let fetchRequest: NSFetchRequest<MBCategory> = MBCategory.fetchRequest()
-            fetchRequest.sortDescriptors = [NSSortDescriptor(key: "parentID", ascending: true)]
+    private func linkCategoriesTogether(managedObjectContext: NSManagedObjectContext) -> Error? {
```
jdwitten, line 84: Nice no more completion block

iOS Client/Store/MBArticlesStore.swift
```diff
@@ -17,184 +17,172 @@ class MBArticlesStore: NSObject {
...
+            // data is an array of Data, where each datum is a serialized array representing a page of results
+            // thus, think of data as an array of results pages
             for jsonData in data {
                 do {
-                    try self.downloadModelsHandler(persistentContainer: persistentContainer,
+                    isNewData = try self.downloadModelsHandler(managedObjectContext: managedObjectContext,
                                                    data: jsonData,
-                                                   deserializeFunc: deserializeFunc,
-                                                   completion: { (isNewData: Bool?, err: Error?) in
-                        serialQueue.async {
```
jdwitten, line 141: What happened to the serial queue? we don't need this anymore?

jdwitten (approved):
Looks good, a couple questions, but nothing to prevent approving.

iOS Client/Store/MBArticlesStore.swift
AlexRamey: It's a subtle thing, but `downloadModelsHandler` used to be asynchronous, so we'd pass it a completion block. Internally, it would use the persistentContainer's performBackgroundTask, which was `async`. The serialQueue was basically the sync point as these background tasks came back. 

After some more thought, I expect that these background tasks were being queued on the same managed object context, meaning they would happen in a serial order and we wouldn't really need this serialQueue to guard against threads colliding into each other. But that was my fear at the time.

Anyway, now downloadModelsHandler is synchronous b/c we are passing in the managedObjectContext. The cool thing I learned is that you don't have to worry so much about threads if you just use the block-based APIs `perform` and `performAndWait` for NSManagedObjectContext. They guarantee that the work happens in the queue designated for the context. 

From the docs:

```
perform(_:) and performAndWait(_:) ensure the block operations are executed on the queue specified for the context. The perform(_:) method returns immediately and the context executes the block methods on its own thread. With the performAndWait(_:) method, the context still executes the block methods on its own thread, but the method doesn’t return until the block is executed.
```

This allowed me to make `downloadModelsHandler` sync, and greatly simplified this function that calls downloadModelsHandler for every page

PR #33, opened 2017-12-28
Feature/podcasts detail

jdwitten:
* Gets us started playing podcasts, just starts playing when you navigate to the page and ends when you leave.
* Need to add play/pause controls


iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+        // Do any additional setup after loading the view.
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        MBStore.sharedStore.subscribe(self)
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillAppear(animated)
```
AlexRamey: viewWillDisappear?

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -127,6 +127,23 @@ class MBClient: NSObject {
         }.resume()
     }
     
+    func getPodcastAudio(for podcast: MBPodcast, completion: @escaping (Data?, Error?) -> Void ) {
```
AlexRamey: It seems like the AVPlayer handles this for us down below. Is this function used?

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+        MBStore.sharedStore.unsubscribe(self)
+    }
+    
+    func configureBackButton() {
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem(title: "Back", style: .done, target: self, action: #selector(self.backToPodcasts(sender:)))
+    }
+    
+    @objc func backToPodcasts(sender: AnyObject) {
+        player = nil
```
AlexRamey: Why set player to nil? Won't the PopCurrentNavigation action deallocate this view controller and, in turn, its fields? Is there a noticeable delay prompting this?

AlexRamey (changes requested):
basically seems fine, just a few comments.

iOS Client/HTTPClient/MBClient.swift
jdwitten: I had this before I realized I could just pass a url. I can remove

iOS Client/PodcastsController/PodcastDetailViewController.swift
jdwitten: nice catch

iOS Client/PodcastsController/PodcastDetailViewController.swift
jdwitten: yeah probably don't need to do this at all

AlexRamey (approved):

PR #34, opened 2017-12-28
Parse devotions

AlexRamey:
Usage: `perl parse_devotions.pl < devotions.txt > iOS\ Client/Data/devotions.json`

I started with the word doc and exported it as plain text. Then I manually replaced smart quotes with normal quotes and trimmed the text before the first devotion date and after the last devotion. The result is the input, devotions.txt.

parse_devotions.pl parses it and outputs devotions.json for our app to use


jdwitten (approved):
Nice! this is sweet, glad we didn't have to do this 365 times

PR #35, opened 2017-12-29
1. Traded memory for speed

AlexRamey:
Thought about embedding child view controllers, but, in the end, all I really wanted was to keep it all in RAM.

This is the simple way


jdwitten (approved):
Nice, I like it. I don't think we will ever get enough articles that this will become a memory issue, but definitely something to keep an eye on

PR #36, opened 2017-12-29
only push safari view controller when user clicks a link

AlexRamey:


jdwitten (commented):
What was going on here that prompted this?

jdwitten (approved):

PR #38, opened 2018-01-03
Feature/podcasts play pause

jdwitten:


iOS Client/Actions/MBActions.swift
```diff
@@ -51,3 +51,9 @@ struct LoadedPodcasts: Action {
 struct SelectedPodcast: Action {
     var podcast: MBPodcast
 }
+
+struct ResumePodcast: Action {
+}
+
+struct PausePodcast: Action {
+}
```
AlexRamey: Just a thought, but I wonder if we could get away with a single action that just toggles it

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -22,6 +23,10 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
     
     let podcastsStore = MBPodcastsStore()
     
+    var player = AVPlayer()
+    var currentPlayingPodcast: MBPodcast?
+    var playerState: PlayerState = .initialized
```
AlexRamey, line 28: interesting, so this is a locally cached substate

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -43,10 +48,45 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
+        if let guid = podcast.guid,
+        let url = URL(string:guid) {
+            let item = AVPlayerItem(url: url)
+            player.replaceCurrentItem(with: item)
+            player.play()
+        }
+    }
+}
+
+enum PlayerState {
```
AlexRamey: do this belong here or in `MBState` ?

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -43,10 +48,45 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
+        playerState = state.podcastsState.player
+        
         guard state.navigationState.selectedTab == .podcasts, let newRoute = state.navigationState.routes[.podcasts] else {
             return
         }
         build(newRoute: newRoute)
         route = newRoute
     }
+    
+    func start(_ podcast: MBPodcast) {
```
AlexRamey: perhaps this function should only accept a string since that's all that it needs

AlexRamey: having it accept an MBPodcast is partly in violation of the `interface segregation principle`. As a justification, think about how we could reuse this function from a context where we didn't have an MBPodcast

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -43,10 +48,45 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
     func newState(state: MBAppState) {
+        
+        switch state.podcastsState.player {
+        case .initialized:
+            break
+        case .playing:
+            if let podcast = state.podcastsState.selectedPodcast, podcast.guid != currentPlayingPodcast?.guid {
+                start(podcast)
+                currentPlayingPodcast = state.podcastsState.selectedPodcast
+            } else if case .paused = playerState {
```
AlexRamey: can we remove the condition and just do `else { player.play() }` ?

AlexRamey: I think it will just have no effect if it is already playing, and it will cover us if somehow the player is in an `error` state

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -84,6 +84,11 @@ func podcastsReducer(action: Action, state: PodcastsState?) -> PodcastsState {
         nextState.podcasts = action.podcasts
     case let action as SelectedPodcast:
         nextState.selectedPodcast = action.podcast
+        nextState.player = .playing
+    case let action as ResumePodcast:
+        nextState.player = .playing
+    case let action as PausePodcast:
+        nextState.player = .paused
     default:
```
AlexRamey: I'm envisioning 
```
case let action as PlayPausePodcast:
    nextState.player = (nextState.player == .playing ? .paused : .playing )
```

iOS Client/State/MBState.swift
```diff
@@ -67,11 +67,13 @@ struct MBDevotionState: DevotionState {
...
     var podcasts: Loaded<[MBPodcast]> { get set }
     var selectedPodcast: MBPodcast? { get set }
+    var player: PlayerState { get set }
 }
 
 struct MBPodcastsState: PodcastsState {
     var podcasts: Loaded<[MBPodcast]> = .initial
     var selectedPodcast: MBPodcast? = nil
+    var player: PlayerState = .initialized
 }
```
AlexRamey, line 78: hmm, maybe it wasn't the best idea on my part to protocol out all this stuff . . . but that's neither here nor there for this PR. I'd be curious to know if yall use protocols at work for your state

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -9,21 +9,62 @@
...
+        playPauseView.bottomAnchor.constraint(equalTo: tabBar.topAnchor).isActive = true
+        playPauseView.isHidden = true
+        
+        playPauseButton = UIButton()
+        playPauseButton.setTitle("Play", for: .normal)
+        playPauseButton.addTarget(self, action: #selector(togglePlayPause(_:)), for: .touchUpInside)
+        playPauseButton.translatesAutoresizingMaskIntoConstraints = false
+        playPauseView.addSubview(playPauseButton)
+        playPauseButton.centerXAnchor.constraint(equalTo: playPauseView.centerXAnchor).isActive = true
+        playPauseButton.centerYAnchor.constraint(equalTo: playPauseView.centerYAnchor).isActive = true
```
AlexRamey, line 41: i'm gonna do the "trust fall" for this code block

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -9,21 +9,62 @@
...
+    }
+    
+    @objc func togglePlayPause(_ sender: UIButton) {
+        let action: Action
+        if playerState == .paused || playerState == .initialized {
```
AlexRamey: how could the player State be `.initialized` here?

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -9,21 +9,62 @@
...
+            action = ResumePodcast()
+        } else {
+            playPauseButton.setTitle("Play", for: .normal)
```
AlexRamey: shouldn't this happen in newState once the state is updated?

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -35,15 +76,19 @@ class MBTabBarController: UITabBarController {
...
-    // In a storyboard-based application, you will often want to do a little preparation before navigation
-    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
-        // Get the new view controller using segue.destinationViewController.
-        // Pass the selected object to the new view controller.
+    
+    func newState(state: MBAppState) {
+        if state.podcastsState.player != .initialized {
+            playPauseView?.isHidden = false
+        }
+
```
AlexRamey, line 77: Nice, do you think this should be hidden on article and/or devotion detail?

AlexRamey (changes requested):
cool spotify action

iOS Client/PodcastsController/PodcastsCoordinator.swift
jdwitten: Yeah basically whether to tell if the player is already playing when a new state comes in. (so we don't have to call player.play() again )

iOS Client/PodcastsController/PodcastsCoordinator.swift
jdwitten: No so I tried that but the podcast will start over on player.play()

iOS Client/PodcastsController/PodcastsCoordinator.swift
jdwitten: of course the classic interface segregation principle

iOS Client/PodcastsController/PodcastsCoordinator.swift
jdwitten: I can put it in MBState, it's probably more appropriate there

iOS Client/Actions/MBActions.swift
jdwitten: Yeah we might be able to get away with just that, good call

iOS Client/State/MBState.swift
jdwitten: yeah protocols here seem a little bit unnecessary, I don't think they're really serving us any purpose. We don't use them on our state

iOS Client/TabController/MBTabBarController.swift
jdwitten: It's just setting up the view for play/pause above the tab bar, nothing too crazy. We can talk about design of this later

iOS Client/TabController/MBTabBarController.swift
jdwitten: Yeah I think I change this in my next branch but I can move it here too

iOS Client/TabController/MBTabBarController.swift
jdwitten: Don't think it ever will be, but for the compiler not to complain about using `action` all the cases need to be covered in the enum

iOS Client/TabController/MBTabBarController.swift
jdwitten: I could envision someone just reading a devotion or article with a podcast in the background and then want to pause it. Open for debate though

iOS Client/Reducers/MBAppReducer.swift
jdwitten: was having some trouble with the compiler with the ternary operator so just went with an if else block

AlexRamey (approved):

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -9,21 +9,55 @@
...
+        playPauseButton.centerYAnchor.constraint(equalTo: playPauseView.centerYAnchor).isActive = true
+    }
+    
+    @objc func togglePlayPause(_ sender: UIButton) {
+        MBStore.sharedStore.dispatch(PlayPausePodcast())
+    }
     
     override func viewDidLoad() {
         super.viewDidLoad()
+         configurePlayPauseView()
```
AlexRamey, line 50: extra whitespace

PR #39, opened 2018-01-05
Feature/podcast detail player

jdwitten:
* Pretty neat stuff here, we've got a player in the podcast detail vc with a play/pause button, a slider to seek, and a label that tracks current position to total seconds.
* Using a timer in the PodcastCoordinator to update the current duration and the slider. 
* Slider intercepts a touch event and then dispatches an action to seek to appropriate second


AlexRamey:
merge that other one, and then I'll take a look here once the diff goes down

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -38,15 +47,77 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
+        }
+    }
+    
+    func seek(to second: Double) {
+        let time = CMTime(seconds: second, preferredTimescale: 1)
+        player.seek(to: time)
+    }
+}
+
+protocol PodcastHandler {
```
AlexRamey: This should be called PodcastDetailViewControllerDelegate ish

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,19 +53,48 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+    }
+    
+    func updateCurrentDuration(current: Double, total: Double ) {
+        totalDuration = total
+        durationLabel.text = "\(formatter.string(from: NSNumber(value: current)) ?? "0") / \(formatter.string(from: NSNumber(value: total)) ?? "0")"
+        durationSlider.setValue(Float(current/(totalDuration ?? 0.0)), animated: true)
     }
 }
+
+protocol PodcastPlayerDelegate: class {
```
AlexRamey: this shouldn't exist

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -9,9 +9,14 @@
 import Foundation
 import UIKit
 import ReSwift
+import AVKit
 
-class PodcastsCoordinator: Coordinator, StoreSubscriber {
+class PodcastsCoordinator: NSObject, Coordinator, StoreSubscriber, AVAudioPlayerDelegate, PodcastHandler {
+    
     var childCoordinators: [Coordinator] = []
+    weak var playerDelegate: PodcastPlayerDelegate? {
```
AlexRamey: this shouldn't exist.

The point of delegation is to create a relationship at runtime between an object that has something to report and another object willing to listen by conforming to some interface.

It makes sense to make the coordinator a delegate of the view controller, where the coordinator is stored as a weak property on the VC to avoid a retain cycle. This way the VC can report events back up to the coordinator. 

However, delegation the other way doesn't make as much sense b/c the coordinator already knows about this view controller. The coordinator should just be able to directly invoke some method on it.

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -38,15 +47,77 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
+            break
+        }
+        
+        // Handle changes in navigation state
         guard state.navigationState.selectedTab == .podcasts, let newRoute = state.navigationState.routes[.podcasts] else {
             return
         }
         build(newRoute: newRoute)
         route = newRoute
+        playerDelegate?.handler = self
```
AlexRamey: confusing. This should be something more like podcastDetailViewController.delegate = self
 

maybe if let vc = navigationController.viewControllers.last as? PodcastDetailViewController {
  vc.delegate = self
}

AlexRamey (changes requested):

iOS Client/PodcastsController/PodcastsCoordinator.swift
jdwitten: Yeah I get what you're saying, the thing is that it's the view controller that needs to listen to messages from the coordinator (specifically the updating of the timestamp on the player). I would normally send this via an action and have the vc get a newState, but didn't want to flood the whole app with an action every second while the a podcast is playing.

I think your last point about the coordinator already knowing about the vc is what I can fix. I don't need another protocol, I can just call the method on the vc directly. I'll try that.

AlexRamey: 👍

AlexRamey (approved):
Nice.

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,19 +53,43 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        case .playing:
+            playPauseButton.setImage(#imageLiteral(resourceName: "pause"), for: .normal)
         }
+        titleLabel.text = state.podcastsState.selectedPodcast?.title
+    }
+    
+    func updateCurrentDuration(current: Double, total: Double ) {
+        totalDuration = total
+        durationLabel.text = "\(formatter.string(from: NSNumber(value: current)) ?? "0") / \(formatter.string(from: NSNumber(value: total)) ?? "0")"
+        durationSlider.setValue(Float(current/(totalDuration ?? 0.0)), animated: true)
```
AlexRamey: divide by 0

AlexRamey (changes requested):

iOS Client/PodcastsController/PodcastDetailViewController.swift
AlexRamey: just use `total` instead of `totalDuration` here, and maybe do a guard to make sure total is not equal to 0 before doing this line

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -8,19 +8,31 @@
...
 class PodcastDetailViewController: UIViewController, StoreSubscriber {
     
-    var podcast: MBPodcast?
-    var player: AVPlayer?
+    var totalDuration: Double?
     
+    @IBOutlet weak var durationSlider: UISlider!
+    @IBOutlet weak var imageView: UIImageView!
     @IBOutlet weak var titleLabel: UILabel!
+    @IBOutlet weak var stackView: UIStackView!
```
AlexRamey, line 20: probably don't need an outlet to this stackView . . . doesn't seem to be used anywhere

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,19 +53,46 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
         MBStore.sharedStore.dispatch(PopCurrentNavigation())
     }
 
+    @IBAction func pressPlayPause(_ sender: Any) {
+        MBStore.sharedStore.dispatch(PlayPausePodcast())
+    }
+    
+    @objc func onSeek(slider: UISlider, event: UIEvent) {
+        if let touchEvent = event.allTouches?.first {
+            let secondToSeekTo = Double(slider.value) * (totalDuration ?? 0.0)
```
AlexRamey, line 62: maybe worth a guard to not even continue if totalDuration is nil

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,19 +53,46 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        case .error, .initialized, .paused, .finished:
+            playPauseButton.setImage(#imageLiteral(resourceName: "play"), for: .normal)
+        case .playing:
+            playPauseButton.setImage(#imageLiteral(resourceName: "pause"), for: .normal)
+        }
+        titleLabel.text = state.podcastsState.selectedPodcast?.title
+    }
+    
+    func updateCurrentDuration(current: Double, total: Double ) {
+        totalDuration = total
```
AlexRamey, line 91: it feels like totalDuration should be set in `newState` or from the coordinator when a new podcast starts playing.  I don't think this function should accept it as a param or set it.

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,19 +53,46 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        durationLabel.text = "\(formatter.string(from: NSNumber(value: current)) ?? "0") / \(formatter.string(from: NSNumber(value: total)) ?? "0")"
+        guard let validTime = totalDuration, validTime > 0 else {
```
AlexRamey, line 93: nice, this should probably go first in the function b/c it will look weird to have "0 / 0" on the label. Does the number formatter display time values? like 0:22 / 3:22 ?

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,19 +53,46 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+            return
         }
+        durationSlider.setValue(Float(current/validTime), animated: true)
```
AlexRamey, line 96: is this line necessary, I'd think the slider would already be in position?

AlexRamey: Hmmm, probably in the seek case but maybe not when a new podcast starts to play . . .

AlexRamey (approved):
Looks like it will work; we can clean it up as we go . . .

PR #40, opened 2018-01-13
Feature/data cleanup job

AlexRamey:
Also put in a refresh control on the table view


iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -47,34 +49,42 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
...
-            if articles.count > 0 {
-                MBStore.sharedStore.dispatch(LoadedArticles(articles: .loaded(data: articles)))
-                return
-            }
-        }
+        let articles = articlesStore.getArticles(managedObjectContext: self.managedObjectContext)
+        MBStore.sharedStore.dispatch(LoadedArticles(articles: .loaded(data: articles)))
         
-        downloadArticleData()
+        if lastUpdateTimestamp < oneDayAgoTimestamp || articles.count == 0 {
```
jdwitten: Is there a reason not to refresh every time viewDidLoad is called. Since this will only really be called if the app is killed and rebooted I would think people would want to see all the latest articles when that happened.

iOS Client/ArticlesController/MBArticlesViewController.swift
```diff
@@ -85,17 +95,26 @@ class MBArticlesViewController: UIViewController, UITableViewDelegate, UITableVi
     func newState(state: MBAppState) {
         switch state.articleState.articles {
         case .initial:
-            break //Do nothing if articles haven't tried to load
+            break
         case .loading:
-            break //Do something here to indicate loading
+            if case Loaded<[MBArticle]>.loading = self.currentState.articles {
```
jdwitten: could just be `if case .loading = self.currentState.articles` I think.

iOS Client/Store/MBArticlesStore.swift
```diff
@@ -183,4 +183,42 @@ class MBArticlesStore: NSObject {
             throw MBDeserializationError.contractMismatch(msg: "unable to cast json object into an array of NSDictionary objects")
         }
     }
+    
+    /***** Data Cleanup Task *****/
+    func deleteOldArticles(managedObjectContext: NSManagedObjectContext, completion: @escaping (Int) -> Void) {
```
jdwitten, line 188: nice looks good

iOS Client/Constants/MBConstants.swift
```diff
@@ -14,5 +14,6 @@ struct MBConstants {
     static let DEFAULTS_KEY_BACKGROUND_APP_REFRESH_ATTEMPT_TIMESTAMP = "DEFAULTS_KEY_BACKGROUND_APP_REFRESH_ATTEMPT_TIMESTAMP"
     static let SECONDS_IN_A_WEEK: Double = 7*24*3600
     static let SECONDS_IN_A_DAY: Double = 24*3600
+    static let MAX_ARTICLES_ON_DEVICE: Int = 200
```
jdwitten: Is there a reason you chose 200?

jdwitten (changes requested):
Just one question I had about when to refresh articles

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: agreed

iOS Client/ArticlesController/MBArticlesViewController.swift
AlexRamey: changed it

iOS Client/Constants/MBConstants.swift
AlexRamey: nope. I was thinking about exposing this to the user through a setting somehow (maybe a low, medium, high setting for how many articles they'd like to persist . . .). There just has to be some cutoff

jdwitten (approved):

PR #44, opened 2018-01-14
Accessibility/font size articles

AlexRamey:


iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -25,24 +25,55 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
...
+            let contentTypeHead = "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\">"
+            let viewPortHead = "<meta name=\"viewport\" content=\"width=device-width,initial-scale=1\">"
+            let cssHead = "<link rel=\"stylesheet\" type=\"text/css\" href=\"MB.css\" />"
+            let head = "<head>\n\(contentTypeHead)\n\(viewPortHead)\n\(cssHead)\n</head>"
+            
+            // <style>
+            let titleFontSize = UIFont.preferredFont(forTextStyle: .headline).pointSize
+            let authorFontSize = UIFont.preferredFont(forTextStyle: .subheadline).pointSize
+            let bodyFontSize = UIFont.preferredFont(forTextStyle: .body).pointSize
+            let style = ""
```
jdwitten, line 61: The only way to adjust the size in the web view is by giving it a new css file? I thought there would be some properties on the webview but if not then we can stick with this

jdwitten (dismissed):

iOS Client/ArticlesController/MBArticleDetailViewController.swift
AlexRamey: yeah . . . I was hoping it would be built-in for the wk web view but that wasn't the case . . . which led to this.

PR #46, opened 2018-01-14
Share article

AlexRamey:
solves #45

### DIFF
--- a/iOS Client/ArticlesController/MBArticleDetailViewController.swift
+++ b/iOS Client/ArticlesController/MBArticleDetailViewController.swift
@@ -25,8 +25,10 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
     
     override func viewDidLoad() {
         super.viewDidLoad()
+        
         configureWebView()
         configureBackButton()
+        configureShareButton()
     }
     
     func configureWebView() {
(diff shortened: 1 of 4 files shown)


iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -61,6 +63,34 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         MBStore.sharedStore.dispatch(PopCurrentNavigation())
     }
     
+    func configureShareButton() {
+        self.navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.action, target: self, action: #selector(self.shareArticle(sender:)))
+    }
+    
+    @objc func shareArticle(sender: AnyObject) {
```
jdwitten, line 70: nice so this is all we do? Just give the activity view controller a link and it's good to go?

jdwitten (approved):

iOS Client/ArticlesController/MBArticleDetailViewController.swift
AlexRamey: yep, it's very simple. Just turning on a built-in thing

PR #47, opened 2018-01-25
1. Changed dates to 2016 so leap year is actually honored when we sch…

AlexRamey:
…edule notifications instead of an error being thrown for that case


jdwitten (approved):
Definitely need to do a deep dive into why these aren't being scheduled.

PR #48, opened 2018-01-29
Feature/all podcast feeds

jdwitten:
* This tries to get all podcasts for all 3 streams and maps them to a new model for displaying with an image.
* Sorts podcasts by date


iOS Client/HTTPClient/MBClient.swift
```diff
@@ -8,6 +8,7 @@
```
AlexRamey, line 8: LOL at these pictures man . . .

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -40,6 +43,20 @@ class MBClient: NSObject {
         case failedPagingRequest(msg: String)
     }
     
+    public enum PodcastStream: String {
+        case pz = "http://pzspodcast.fireside.fm/rss"
```
AlexRamey: can we reuse the fields defined above?

AlexRamey: I notice there's an inconsistency with http vs https for mockingCast. It should be https, same goes for pz

iOS Client/Info.plist
```diff
@@ -20,6 +20,39 @@
 	<string>1</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSExceptionDomains</key>
+		<dict>
+			<key>fireside.fm</key>
```
AlexRamey: I don't think this is necessary. Fireside supports `https` so we can just add that to the urls defined in our client above

iOS Client/Info.plist
```diff
@@ -20,6 +20,39 @@
...
+			<dict>
+				<key>NSIncludesSubdomains</key>
+				<true/>
+				<key>NSTemporaryExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+				<key>NSTemporaryExceptionMinimumTLSVersion</key>
+				<string>TLSv1.1</string>
+			</dict>
+		</dict>
+	</dict>
```
AlexRamey: This file seems kind of janked up with regard to the spacing. Since mbird and youtube were already here, an we don't need fireside, you can just revert this file

iOS Client/HTTPClient/MBClient.swift
AlexRamey: https works for the fireside endpoints, we should add https to our url constants for fireside ones

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -40,6 +43,20 @@ class MBClient: NSObject {
...
+        case mockingCast = "http://themockingcast.fireside.fm/rss"
+        case mockingPulpit = "http://www.mbird.com/feed/podcast/"
+        
+        var image: UIImage {
+            switch self {
+            case .pz: return #imageLiteral(resourceName: "pzcast")
+            case .mockingPulpit: return #imageLiteral(resourceName: "mockingpulpit")
+            case .mockingCast: return #imageLiteral(resourceName: "mockingcast")
+            }
+        }
```
AlexRamey: nice

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -109,22 +126,21 @@ class MBClient: NSObject {
...
-                completion(nil, e)
-            } else if let response = data {
-                completion(response, nil)
-            } else {
-                completion(nil, NetworkRequestError.networkError(msg: "did not receive a response"))
-            }
-        }.resume()
+        let request = URLRequest(url: url)
+        
+        return self.session.dataTask(with: request).asDataAndResponse().then { (data: Data, resp: URLResponse?) -> [MBPodcast] in
```
AlexRamey: very cool

iOS Client/Models/MBPodcast.swift
```diff
@@ -18,3 +19,14 @@ struct MBPodcast: Codable, Detailable {
...
+struct DisplayablePodcast: Detailable {
+    let author: String?
+    let duration: String?
+    let guid: String?
+    let image: UIImage
+    let keywords: String?
+    let summary: String?
+    let pubDate: Date
+    let title: String?
+}
```
AlexRamey, line 32: hmmmm, we need a DRYer way to do this . . . Maybe just make the image's type `Data` instead of UIImage (since `Data` is codable) and keep this one struct

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,14 +53,48 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        case .playing:
+            playPauseButton.setImage(#imageLiteral(resourceName: "pause"), for: .normal)
+        }
+        titleLabel.text = state.podcastsState.selectedPodcast?.title
+    }
+    
+    func updateCurrentDuration(current: Double, total: Double ) {
+        totalDuration = total
+        durationLabel.text = "\(formatter.string(from: NSNumber(value: current)) ?? "0") / \(formatter.string(from: NSNumber(value: total)) ?? "0")"
+        durationSlider.setValue(Float(current/(totalDuration ?? 0.0)), animated: true)
```
AlexRamey: this branch will need to be rebased onto master once you merge the other PR

AlexRamey: (or possibly merged . . .)

AlexRamey (changes requested):

iOS Client/HTTPClient/MBClient.swift
```diff
@@ -40,6 +40,20 @@ class MBClient: NSObject {
...
+        case mockingPulpit = "http://www.mbird.com/feed/podcast/"
+        
+        var imageName: String {
+            switch self {
+            case .pz: return "pzcast"
+            case .mockingPulpit: return "mockingpulpit"
+            case .mockingCast: return "mockingcast"
+            }
+        }
+    }
```
AlexRamey, line 55: nice, much cleaner

iOS Client/Models/MBPodcast.swift
AlexRamey: dude, these are like identical models, anyway we can get away with just having one? 

Looks like pubDate is a date on DisplayablePodcast and that's the only difference, could just be a lazy property on the first struct that gets derived from the string date

iOS Client/PodcastsController/MBPodcastsViewController.swift
```diff
@@ -48,9 +48,9 @@ class MBPodcastsViewController: UIViewController, UITableViewDataSource, UITable
         // swiftlint:enable force_cast
         if indexPath.row < podcasts.count {
```
AlexRamey, line 49: this is fine, but in general I usually live on the edge in the table view methods. I'd rather throw in an `!` than handle some error case that will never happen

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -8,19 +8,31 @@
...
+    @IBOutlet weak var imageView: UIImageView!
     @IBOutlet weak var titleLabel: UILabel!
+    @IBOutlet weak var stackView: UIStackView!
+    @IBOutlet weak var playPauseButton: UIButton!
+    @IBOutlet weak var durationLabel: UILabel!
+    
+    let formatter: NumberFormatter = {
+        let f = NumberFormatter()
+        return f
+    }()
```
AlexRamey: nice, very performant 👍

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,14 +53,48 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+            playPauseButton.setImage(#imageLiteral(resourceName: "play"), for: .normal)
+        case .playing:
+            playPauseButton.setImage(#imageLiteral(resourceName: "pause"), for: .normal)
+        }
+        titleLabel.text = state.podcastsState.selectedPodcast?.title
+    }
+    
+    func updateCurrentDuration(current: Double, total: Double ) {
+        totalDuration = total
+        durationLabel.text = "\(formatter.string(from: NSNumber(value: current)) ?? "0") / \(formatter.string(from: NSNumber(value: total)) ?? "0")"
```
AlexRamey: these numbers will need to be formatted in mm:ss format, maybe the NumberFormatter can help . . .?

iOS Client/PodcastsController/PodcastDetailViewController.swift
AlexRamey: hmmmmmm, we should make sure `totalDuration != 0`

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -41,14 +53,48 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        durationSlider.setValue(Float(current/(totalDuration ?? 0.0)), animated: true)
     }
 }
+
+protocol PodcastPlayerDelegate: class {
+    func updateCurrentDuration(current: Double, total: Double )
+    var handler: PodcastHandler? { get set }
```
AlexRamey: I don't get why this protocol exists? I thought we had addressed this earlier . . .

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -81,4 +104,27 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
+    // MARK: - AVAudioPlayerDelegate
+    func audioPlayerDidFinishPlaying(_ player: AVAudioPlayer, successfully flag: Bool) {
+        if flag {
+            MBStore.sharedStore.dispatch(FinishedPodcast())
+        } else {
+            MBStore.sharedStore.dispatch(PodcastError())
+        }
+    }
+    
+    func seek(to second: Double) {
```
AlexRamey: this should be explicitly marked as the `PodcastHandler` method, which should really be called something like `PodcastViewControllerDelegate` if you decide to stick with delegation for the callback. I would say it's a no brainer, but with UDF in the equation I could see a desire to fire an action and let that come back down. Up to you. Delegation seems to be much more direct

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -10,9 +10,14 @@ import Foundation
...
 import ReSwift
 import AVKit
+import PromiseKit
 
-class PodcastsCoordinator: Coordinator, StoreSubscriber {
+class PodcastsCoordinator: NSObject, Coordinator, StoreSubscriber, AVAudioPlayerDelegate, PodcastHandler {
+    
     var childCoordinators: [Coordinator] = []
+    weak var playerDelegate: PodcastPlayerDelegate? {
+        return navigationController.viewControllers.last as? PodcastPlayerDelegate
```
AlexRamey: this shouldn't exist

AlexRamey: as discussed previously

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -81,4 +104,27 @@ class PodcastsCoordinator: Coordinator, StoreSubscriber {
...
+        }
+    }
+    
+    func seek(to second: Double) {
+        let time = CMTime(seconds: second, preferredTimescale: 1)
+        player.seek(to: time)
+    }
+}
+
+protocol PodcastHandler {
```
AlexRamey: `PodcastHandler` and `PodcastPlayerDelegate` protocol definitions should be swapped in terms of placement.

Think about how table view works. Right next to table view, they define `TableViewDataSource` and `TableViewDelegate` protocols. And the table view has properties of those types. Then outsiders can conform to those protocols and interface with table view.

In this case, the coordinator has a property of type `PodcastPlayerDelegate`. That protocol should be defined in this file. Then, the Podcast View Controller could implement it. Similarly, the Podcast View controller file should hold the definition of `PodcastHandler`. 

However, again, all of this is a moot point b/c we shouldn't have two-way delegation going on. This coordinator should be calling functions directly on the view controller. We don't need protocols for that direction. However, if you want to setup a callback scheme from the view controller back up to the coordinator, then delegation is the move for that direction. You'd define some delegate protocol in the view controller file and then conform to it in the coordinator file and directly set the view controller's delegate to self (the coordinator).

Think of delegation as a solution to when you need a callback. 

Remember, we know all about the table view, so we can tell it to do whatever we want without delegation. But the table view has no idea who we are, so we set its delegate property in order to receive callbacks. All the table view knows is that we conform to its delegate protocol.

With these thoughts in mind, I think this code can be made much simpler, which I thought was done earlier on b/c I'm pretty sure we had this discussion, but maybe GitHub is being weird.

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -8,38 +8,51 @@
...
-                } catch let error {
-                    completion(nil, error)
+    func syncPodcasts() -> Promise<[DisplayablePodcast]> {
+        let streams: [MBClient.PodcastStream] = [.pz, .mockingPulpit, .mockingCast]
+        let requests = streams.map{ self.client.getPodcasts(for: $0)}
+        return firstly {
+            when(resolved: requests)
+        }.then { responses -> Promise<[DisplayablePodcast]> in
+            var podcasts: [DisplayablePodcast] = []
+            for (indx, response) in responses.enumerated() {
```
AlexRamey: `indx` is a bit strange. I've seen `idx` before . . . not a big deal

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -8,38 +8,51 @@
...
+                if case .fulfilled(let newCasts) = response {
+                    let displayCasts = newCasts.flatMap { podcast -> DisplayablePodcast? in
+                        guard let dateString = podcast.pubDate, let date = self.dateFormatter.date(from: dateString) else {
+                            return nil
+                        }
+                        return DisplayablePodcast(author: podcast.author,
```
AlexRamey: Hmm, I go back and forth. There may be some value in splitting our DTO (data transfer object) model from our domain model . . .

iOS Client/Models/MBPodcast.swift
AlexRamey: Hmm, I'm actually not sure. Maybe if we rename MBPodcast to be called `PodcastDTO` (data transfer object) and rename DisplayablePodcast to be `Podcast`, I'll be okay with this

AlexRamey (changes requested):

iOS Client/PodcastsController/PodcastsCoordinator.swift
jdwitten: Yeah man so I just never merged those changes back into this branch, but it should be good to go now if you want to give it a rereview. Not sure which of these comments are still relevant so I would say give it a good look over the whole thing

iOS Client/Store/MBPodcastsStore.swift
jdwitten: My original intent for splitting it was for the images, but now that we are just storing the name of the image I think it might be ok to consolidate to one model

iOS Client/PodcastsController/MBPodcastsViewController.swift
```diff
@@ -33,7 +33,7 @@ class MBPodcastsViewController: UIViewController, UITableViewDataSource, UITable
         MBStore.sharedStore.unsubscribe(self)
     }
     
-    // MARK - UITableViewDataSource
+    // MARK: - UITableViewDataSource
     func numberOfSections(in tableView: UITableView) -> Int {
```
AlexRamey, line 37: nice

iOS Client/PodcastsController/MBPodcastsViewController.swift
```diff
@@ -59,7 +59,7 @@ class MBPodcastsViewController: UIViewController, UITableViewDataSource, UITable
         MBStore.sharedStore.dispatch(SelectedPodcast(podcast: podcasts[indexPath.row]))
     }
     
-    // MARK - StoreSubscriber
+    // MARK: - StoreSubscriber
```
AlexRamey, line 62: nice

iOS Client/PodcastsController/PodcastTableViewCell.swift
```diff
@@ -10,6 +10,8 @@ import UIKit
 
 class PodcastTableViewCell: UITableViewCell {
 
+
```
AlexRamey, line 13: whitespace, doesn't really matter

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -8,38 +8,51 @@
...
+import PromiseKit
 
 
 class MBPodcastsStore {
     
     let client: MBClient
     let fileHelper: FileHelper
+    let dateFormatter: DateFormatter = {
+        let d = DateFormatter()
+        d.dateFormat = "E, d MMM yyyy HH:mm:ss Z"
```
AlexRamey, line 20: this is some intense s*** man, super BA looking.

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -8,38 +8,51 @@
...
-                    let podcasts = xmlParserDelegate.podcasts
-                    try self.fileHelper.save(podcasts, forPath: "podcasts")
-                    completion(podcasts, nil)
-                } catch let error {
-                    completion(nil, error)
+    func syncPodcasts() -> Promise<[DisplayablePodcast]> {
+        let streams: [MBClient.PodcastStream] = [.pz, .mockingPulpit, .mockingCast]
+        let requests = streams.map{ self.client.getPodcasts(for: $0)}
+        return firstly {
+            when(resolved: requests)
```
AlexRamey: very classy use of `when`

iOS Client/Store/XMLParsingDelegate.swift
```diff
@@ -19,6 +19,14 @@ class PodcastXMLParsingDelegate: NSObject, XMLParserDelegate {
         self.foundCharacters += string
     }
     
+    func parser(_ parser: XMLParser, didStartElement elementName: String, namespaceURI: String?, qualifiedName qName: String?, attributes attributeDict: [String : String]) {
+        if elementName == "enclosure" {
+            if let url = attributeDict["url"] {
```
AlexRamey, line 24: you can one line this with `if elementName == "enclosure", let url = attributeDict["url"] { . . . }` I think . . . Certainly can do it if you're willing to put the let clause first

AlexRamey: Question: What's the point of this? Wouldn't this just get overwritten below by the `guid` case of `didEndElement`?

iOS Client/Models/MBPodcast.swift
```diff
@@ -7,8 +7,9 @@
 //
 
 import Foundation
+import UIKit
 
-struct MBPodcast: Codable, Detailable {
+struct MBPodcast: Codable {
```
AlexRamey: I like having two different models.

I think we should call this one `PodcastDTO` and `DisplayablePodcast` can be called `Podcast`.

I started this project by prefixing everything with `MB` b/c you had to use a custom prefix (usually 2 or 3 letters) in Objective-C world to avoid namespace collisions, but in swift that is no longer the case . . . 🤦‍♂️

AlexRamey (changes requested):
Looks great.

My biggest critiques are
1) Rename MBPodcast to PodcastDTO and DisplayablePodcast to Podcast
2) Make sure the XML parse logic is sound

iOS Client/Store/XMLParsingDelegate.swift
jdwitten: Yeah so this is a little artifact that the rss feeds are not structured the same way. 2 of the feeds have the url as fields on the xml element, and one has its own element called `guid` with the url. So the feeds that have `enclosure` don't have `guid` so they don't get overwritten. I think we might need to do some reexamination of this logic but I think it's good to get in for now

AlexRamey: ah okay, seems reasonable. And I see you only have access to this `attributeDict` on the `didStartElement` delegate method.

AlexRamey (approved):
nice, I feel good about having a DTO model and a domain model. This is what we're trending toward on lidl

PR #49, opened 2018-01-30
Notification Enhancements

AlexRamey:
0. Explicitly registered devotions coordinator as delegate to the notification center
1. Schedule only up to 50 notifications at once (and look to advance the window every time the app launches)
2. Added foreground notification handling to still show the notification when the app is in the foreground

### DIFF
--- a/iOS Client/DevotionsController/DevotionsCoordinator.swift
+++ b/iOS Client/DevotionsController/DevotionsCoordinator.swift
@@ -128,10 +128,14 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber, UNUserNotifi
 
 extension Date
 {
+    static let ddMMFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "MM-dd"
+        return formatter
+    }()
+
     func toMMddString() -> String
     {
-        let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "MM-dd"
-        return dateFormatter.string(from: self)
+        return Date.ddMMFormatter.string(from: self)
     }
 }
--- a/iOS Client/DevotionsController/DevotionsCoordinator.swift
+++ b/iOS Client/DevotionsController/DevotionsCoordinator.swift
@@ -28,6 +28,7 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber, UNUserNotifi
         let devotionsController = MBDevotionsViewController.instantiateFromStoryboard()
         navigationController.pushViewController(devotionsController, animated: true)
         MBStore.sharedStore.subscribe(self)
+        UNUserNotificationCenter.current().delegate = self
         let devotions = devotionsStore.getDevotions()
         if devotions.count == 0 {
             devotionsStore.syncDevotions { syncedDevotions, error in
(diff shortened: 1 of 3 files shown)


iOS Client/Models/MBDevotion.swift
```diff
@@ -65,13 +68,15 @@ struct LoadedDevotion: Codable, Detailable {
         return dateComponents
     }
     
-    private static let calendar: NSCalendar? = {
-        return NSCalendar.init(calendarIdentifier: NSCalendar.Identifier.gregorian)
-    }()
+    var dateAsMMdd: String {
+        return String(self.date.split(separator: "-", maxSplits: 1, omittingEmptySubsequences: true)[1])
+    }
```
AlexRamey, line 73: I re-organized this file, but only addition was adding this lazy property

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -36,45 +37,56 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber, UNUserNotifi
...
             }
-            let trigger = UNCalendarNotificationTrigger(dateMatching: dateComponents, repeats: true)
-            let content = DevotionNotificationContent(devotion: devotion)
-            let request = UNNotificationRequest(identifier: UUID().uuidString, content: content, trigger: trigger)
-            center.add(request)
+            
+            var i = startIndex
+            var outstandingNotifications: Int = 0
+            
+            while outstandingNotifications < MBConstants.DEVOTION_NOTIFICATION_WINDOW_SIZE {
```
jdwitten, line 74: Was trying to think of a way to make this more swifty, but I think it gets the job done

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -89,11 +101,37 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber, UNUserNotifi
...
         }
-        MBStore.sharedStore.dispatch(LoadedDevotions(devotions: .loaded(data: devotions)))
+    }
+}
+
+extension Date
+{
+    func toMMddString() -> String
+    {
+        let dateFormatter = DateFormatter()
```
jdwitten: Apparently DateFormatters are actually expensive to initialize so I would make this a static or lazy property. Don't think it really will have an effect but we are looping over all the devotions and initializing a date formatter every time so it might.

jdwitten (changes requested):
Just a small change on the date formatter

iOS Client/DevotionsController/DevotionsCoordinator.swift
AlexRamey: I'M_ALL_ABOUT_THAT_C++ lol

iOS Client/DevotionsController/DevotionsCoordinator.swift
AlexRamey: reasonable

iOS Client/DevotionsController/DevotionsCoordinator.swift
```diff
@@ -89,11 +101,41 @@ class DevotionsCoordinator: NSObject, Coordinator, StoreSubscriber, UNUserNotifi
...
+        if let devotion = devotions.first(where: {$0.dateAsMMdd == Date().toMMddString()}) {
             MBStore.sharedStore.dispatch(SelectedDevotion(devotion: devotion))
         }
-        MBStore.sharedStore.dispatch(LoadedDevotions(devotions: .loaded(data: devotions)))
+    }
+}
+
+extension Date
+{
+    static let ddMMFormatter: DateFormatter = {
```
jdwitten, line 131: nice looks good

jdwitten (approved):

Issue #43, opened 2018-01-13
Bookmarks


AlexRamey:
Take care not to let these get deleted by the delete job (should just need to add a predicate to the fetch requests in that function to ignore bookmarked articles)

PR #52, opened 2018-01-31
1. Added bookmarks feature

AlexRamey:
solves #43

### DIFF
--- a/iOS Client/ArticlesController/ArticlesCoordinator.swift
+++ b/iOS Client/ArticlesController/ArticlesCoordinator.swift
@@ -16,7 +16,7 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
     var childCoordinators: [Coordinator] = []
     var route: [RouteComponent] = [.base]
     var tab: Tab = .articles
-    var overlay: URL? = nil
+    var overlay: URL?
     let managedObjectContext: NSManagedObjectContext
     
     var rootViewController: UIViewController {
@@ -64,6 +64,5 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
     // MARK: - SafariViewControllerDelegate
     func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
         MBStore.sharedStore.dispatch(SelectedArticleLink(url: nil))
-        self.overlay = nil
     }
 }
--- a/iOS Client/ArticlesController/ArticlesCoordinator.swift
+++ b/iOS Client/ArticlesController/ArticlesCoordinator.swift
@@ -12,7 +12,7 @@ import ReSwift
 import SafariServices
 import CoreData
 
-class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationControllerDelegate, SafariDisplayer {
+class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationControllerDelegate, SafariDisplayer, SFSafariViewControllerDelegate {
     var childCoordinators: [Coordinator] = []
     var route: [RouteComponent] = [.base]
     var tab: Tab = .articles
@@ -49,6 +49,11 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
         build(newRoute: newRoute)
         route = newRoute
         
-        self.displaySafariVC(forURL: state.navigationState.safariOverlays[.articles]!)
+        self.displaySafariVC(forURL: state.navigationState.safariOverlays[.articles].flatMap {return $0}, withDelegate: self)
+    }
+    
+    // MARK: - SafariViewControllerDelegate
+    func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
+        MBStore.sharedStore.dispatch(SelectedArticleLink(url: nil))
     }
 }
--- a/iOS Client/ArticlesController/ArticlesCoordinator.swift
+++ b/iOS Client/ArticlesController/ArticlesCoordinator.swift
@@ -12,7 +12,7 @@ import ReSwift
 import SafariServices
 import CoreData
 
-class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationControllerDelegate, SFSafariViewControllerDelegate {
+class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationControllerDelegate, SafariDisplayer {
     var childCoordinators: [Coordinator] = []
     var route: [RouteComponent] = [.base]
     var tab: Tab = .articles
@@ -23,7 +23,7 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
         return self.navigationController
     }
     
-    private lazy var navigationController: UINavigationController = {
+    lazy var navigationController: UINavigationController = {
         return UINavigationController()
     }()
     
@@ -49,20 +49,6 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
         build(newRoute: newRoute)
         route = newRoute
         
-        let articlesSafariOverlay: URL? = state.navigationState.safariOverlays[.articles]!
-        if self.overlay == nil, let overlay = articlesSafariOverlay {
-            self.overlay = overlay
-            let overlayVC = SFSafariViewController(url: overlay)
-            overlayVC.delegate = self
-            self.navigationController.present(overlayVC, animated: true, completion: nil)
-        } else if self.overlay != nil, articlesSafariOverlay == nil {
-            self.overlay = nil
-            self.navigationController.dismiss(animated: true, completion: nil)
-        }
-    }
-    
-    // MARK: - SafariViewControllerDelegate
-    func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
-        MBStore.sharedStore.dispatch(SelectedArticleLink(url: nil))
+        self.displaySafariVC(forURL: state.navigationState.safariOverlays[.articles]!)
     }
 }
--- a/iOS Client/ArticlesController/MBArticleDetailViewController.swift
+++ b/iOS Client/ArticlesController/MBArticleDetailViewController.swift
@@ -33,7 +33,7 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         
         configureWebView()
         configureBackButton()
-        configureShareButton()
+        configureRightBarButtons()
         
         NotificationCenter.default.addObserver(self,
                                                selector: #selector(contentSizeDidChange(_:)),
@@ -92,8 +92,18 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         MBStore.sharedStore.dispatch(PopCurrentNavigation())
     }
     
-    func configureShareButton() {
-        self.navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.action, target: self, action: #selector(self.shareArticle(sender:)))
+    func configureRightBarButtons() {
+        let bookmarkItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.bookmarks, target: self, action: #selector(self.bookmarkArticle(sender:)))
+        
+        let shareItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.action, target: self, action: #selector(self.shareArticle(sender:)))
+        
+        var items = [shareItem]
+        if self.tabBarController?.selectedIndex != 1 {
+            // if not on bookmarks tab, show bookmark item
+            items.append(bookmarkItem)
+        }
+        
+        self.navigationItem.rightBarButtonItems = items
     }
     
     @objc func shareArticle(sender: AnyObject) {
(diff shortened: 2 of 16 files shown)


iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -92,8 +92,18 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
...
     
-    func configureShareButton() {
-        self.navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.action, target: self, action: #selector(self.shareArticle(sender:)))
+    func configureRightBarButtons() {
+        let bookmarkItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.bookmarks, target: self, action: #selector(self.bookmarkArticle(sender:)))
+        
+        let shareItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonSystemItem.action, target: self, action: #selector(self.shareArticle(sender:)))
+        
+        var items = [shareItem]
+        if self.tabBarController?.selectedIndex != 1 {
```
AlexRamey: this is brittle, do something more clever

AlexRamey: could accept tab as a parameter to the VC's initializer

iOS Client/BookmarksController/BookmarksCoordinator.swift
```diff
@@ -23,20 +26,42 @@ class BookmarksCoordinator: NSObject, Coordinator, StoreSubscriber {
...
+            self.navigationController.present(overlayVC, animated: true, completion: nil)
+        } else if self.overlay != nil, bookmarksSafariOverlay == nil {
+            self.overlay = nil
+            self.navigationController.dismiss(animated: true, completion: nil)
+        }
+    }
+    
+    // MARK: - SafariViewControllerDelegate
+    func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
+        MBStore.sharedStore.dispatch(SelectedArticleLink(url: nil))
```
AlexRamey, line 56: This is a copy paste job from Articles Coordinator, do something more clever

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -27,12 +27,20 @@ func navigationReducer(action: Action, state: NavigationState?) -> NavigationSta
     case let action as NavigationActionSwitchTab:
         nextState.selectedTab = action.tab
     case let action as SelectedArticle:
-        nextState.routes[.articles]?.append(.detail(item: action.article))
+        if nextState.selectedTab == .articles {
+           nextState.routes[.articles]?.append(.detail(item: action.article))
+        } else if nextState.selectedTab == .bookmarks {
+            nextState.routes[.bookmarks]?.append(.detail(item: action.article))
+        }
```
AlexRamey: this can be a one-liner

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -27,12 +27,20 @@ func navigationReducer(action: Action, state: NavigationState?) -> NavigationSta
...
     case let action as SelectedDevotion:
         nextState.selectedTab = .devotions
         nextState.routes[.devotions]?.append(.detail(item: action.devotion))
     case let action as SelectedArticleLink:
-        nextState.safariOverlays[.articles] = action.url
+        if nextState.selectedTab == .articles {
+            nextState.safariOverlays[.articles] = action.url
+        } else if nextState.selectedTab == .bookmarks {
+            nextState.safariOverlays[.bookmarks] = action.url
+        }
```
AlexRamey: this can be a one-liner

iOS Client/Store/MBArticlesStore.swift
```diff
@@ -187,6 +187,8 @@ class MBArticlesStore: NSObject {
     /***** Data Cleanup Task *****/
     func deleteOldArticles(managedObjectContext: NSManagedObjectContext, completion: @escaping (Int) -> Void) {
         let fetchRequest = NSFetchRequest<NSManagedObject>(entityName: MBArticle.entityName)
+        fetchRequest.predicate = NSPredicate(format: "isBookmarked == %@", NSNumber(value: false))
+
```
AlexRamey, line 191: nice 👍

AlexRamey (commented):
I request changes of myself

iOS Client/ArticlesController/MBArticleDetailViewController.swift
AlexRamey: addressed by comparing `!= Tab.bookmarks.rawValue`

iOS Client/Reducers/MBAppReducer.swift
AlexRamey: addressed

iOS Client/Reducers/MBAppReducer.swift
AlexRamey: addressed

iOS Client/BookmarksController/BookmarksCoordinator.swift
AlexRamey: addressed with `SafariDisplayer` protocol and extension methods

AlexRamey:
comments addressed, awaiting additional feedback from jwitt

iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -120,6 +130,22 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
         self.present(activityViewController, animated: true, completion: nil)
     }
     
+    @objc func bookmarkArticle(sender: AnyObject) {
+        self.selectedArticle?.isBookmarked = true
+        
+        var message = ""
+        do {
+            try self.selectedArticle?.managedObjectContext?.save()
```
jdwitten, line 138: Nice my dude a little core data action

iOS Client/ArticlesController/MBArticleDetailViewController.swift
```diff
@@ -120,6 +130,22 @@ class MBArticleDetailViewController: UIViewController, WKNavigationDelegate {
...
+            message = "Bookmarked"
+        } catch {
+            message = error.localizedDescription
+        }
+        
+        let alert = UIAlertController(title: "Done", message: message, preferredStyle: UIAlertControllerStyle.alert)
```
jdwitten, line 144: Does this make the two options on the alert "Done" and "Dismiss"? Because I feel like there should just be one of those, they imply doing the same thing.

iOS Client/BookmarksController/MBBookmarksViewController.swift
```diff
@@ -18,24 +26,122 @@ class MBBookmarksViewController: UIViewController {
...
     override func viewDidLoad() {
         super.viewDidLoad()
-
-        // Do any additional setup after loading the view.
+        self.title = "Bookmarks"
+        self.isEditing = false
+        self.navigationItem.leftBarButtonItem = editButtonItem
+        
+        let fetchRequest: NSFetchRequest<MBArticle> = MBArticle.fetchRequest()
+        fetchRequest.predicate = NSPredicate(format: "isBookmarked == %@", NSNumber(value: true))
```
jdwitten, line 34: sweet dude nice and easy

iOS Client/BookmarksController/MBBookmarksViewController.swift
```diff
@@ -18,24 +26,122 @@ class MBBookmarksViewController: UIViewController {
...
         // Dispose of any resources that can be recreated.
     }
-    
+}
 
-    /*
-    // MARK: - Navigation
+// MARK: - Internal
+extension MBBookmarksViewController {
+    func configure(cell: UITableViewCell, for indexPath: IndexPath) {
```
jdwitten, line 58: Can this be a method on the cell instead of on `MBBookmarksViewController` in case that cell ever has to be reused in a different context

iOS Client/BookmarksController/MBBookmarksViewController.swift
```diff
@@ -18,24 +26,122 @@ class MBBookmarksViewController: UIViewController {
...
-        // Get the new view controller using segue.destinationViewController.
-        // Pass the selected object to the new view controller.
+// MARK: - NSFetchedResultsControllerDelegate
+extension MBBookmarksViewController: NSFetchedResultsControllerDelegate {
+    func controllerWillChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
+        tableView.beginUpdates()
     }
-    */
+    
+    func controller(_ controller: NSFetchedResultsController<NSFetchRequestResult>, didChange anObject: Any, at indexPath: IndexPath?, for type: NSFetchedResultsChangeType, newIndexPath: IndexPath?) {
```
jdwitten, line 71: I'll be honest not sure how this FetchResultsController works so I'm going to trust you got this one

iOS Client/ArticlesController/ArticlesCoordinator.swift
```diff
@@ -49,21 +49,6 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
...
-            self.overlay = nil
-            self.navigationController.dismiss(animated: true, completion: nil)
-        }
-    }
-    
-    // MARK: - SafariViewControllerDelegate
-    func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
-        MBStore.sharedStore.dispatch(SelectedArticleLink(url: nil))
-        self.overlay = nil
+        self.displaySafariVC(forURL: state.navigationState.safariOverlays[.articles]!)
```
jdwitten: How do you know that this is not going to be nil?

jdwitten (changes requested):
Looks mostly good, just some questions and then one change on the configure method on the table view cell

iOS Client/ArticlesController/ArticlesCoordinator.swift
AlexRamey: I invite it to be `nil`. The method will pop that overlay if there currently is one and the passed url is nil.

iOS Client/ArticlesController/MBArticleDetailViewController.swift
AlexRamey: simulator screen shot - iphone x - 2018-02-05 at 21 29 39

Maybe later, for success, we could just do a checkbox or a bookmarking  animation in the nav bar

iOS Client/BookmarksController/MBBookmarksViewController.swift
AlexRamey: ah yeah dude, I copied all this from the book

iOS Client/BookmarksController/MBBookmarksViewController.swift
AlexRamey: Hmm, u make a fair point. I'll move it over.

AlexRamey: addressed

iOS Client/ArticlesController/MBArticleDetailViewController.swift
jdwitten: Cool gotcha the "Done" is the title

iOS Client/ArticlesController/ArticlesCoordinator.swift
jdwitten: Yeah but you explicitly unwrap `state.navigationState.safariOverlays[.articles]` could that ever be nil and then crash the app. I'm not sure if it ever could be, just verifying you considered that

AlexRamey: ah yeah, I think I was thinking that key lookups in dictionaries return optionals, and this is a dictionary of `[Tab : URL?]`, so I thought this would return a `URL??` and I would force unwrap it to a single `URL?` but not actually 100% sure this is how it works . . .

Lemme playground

AlexRamey:
```swift
//: Playground - noun: a place where people can play

import UIKit

var str = "Hello, playground"

var names: [Int: String?] = [:]

func getOptional(str: String?) -> String {
    if let val = str {
        return val
    } else {
        return "nil!"
    }
}

names[0] = "Ricky"
names[1] = "Bobby"

let bobby = getOptional(str: names[1]!)                     // "Bobby"
let none = getOptional(str: names[2]?.flatMap {return $0})  // "nil!"

// getOptional(str: names[2]!) // CRASH
// getOptional(str: names[0]) // ERROR: Value of type String?? not compatible

```

AlexRamey: Good catch, it seems like `!` unwraps all the way. I thought it would just strip the outer-most optional.

I think I will use `flatMap` instead

AlexRamey: I also think I screwed something up with the safari vc delegate, b/c it's not getting called since I moved it into the safari displayer protocol extension.

will investigate tmr . . .

AlexRamey: addressed. Had to move the safari vc delegate behavior back into the coordinators; something not working about doing protocol conformance implementation from another protocol

AlexRamey: Here's some more interesting code: 

```swift
var maybeMaybeOne: Int?? = .some(.some(5))
var maybeOne: Int? = maybeMaybeOne!
var one: Int = maybeMaybeOne!!

var maybeMaybeTwo: Int?? = .some(nil)
var maybeTwo: Int? = maybeMaybeTwo!
// var two: Int = maybeMaybeTwo!! CRASH: Unexpectedly found nil
```

AlexRamey: It seems that ! indeed only strips the outermost ?.

The thing is, a double optional could be `.some(nil)`, in which case a force unwrap crashes, or it could be `.some(.some(nil))`, in which case unwrapping once is fine. 

In the dictionary case above, it's going to be `.some(nil)` or `.some(.some(URL))`. In hindsight, the dictionary should probably not be to an optional URL but to a normal URL instead

iOS Client/ArticlesController/ArticlesCoordinator.swift
```diff
@@ -49,21 +49,11 @@ class ArticlesCoordinator: NSObject, Coordinator, StoreSubscriber, UINavigationC
...
-        if self.overlay == nil, let overlay = articlesSafariOverlay {
-            self.overlay = overlay
-            let overlayVC = SFSafariViewController(url: overlay)
-            overlayVC.delegate = self
-            self.navigationController.present(overlayVC, animated: true, completion: nil)
-        } else if self.overlay != nil, articlesSafariOverlay == nil {
-            self.overlay = nil
-            self.navigationController.dismiss(animated: true, completion: nil)
-        }
+        self.displaySafariVC(forURL: state.navigationState.safariOverlays[.articles].flatMap {return $0}, withDelegate: self)
```
jdwitten, line 52: Nice dude I like it, much better than `!`

jdwitten (approved):
Looks good man

PR #53, opened 2018-02-03
dropped in calendar component for better devotions navigation

AlexRamey:
Pure cash.

I will change the target to master once feature/bookmarks is approved and merged


AlexRamey:
simulator screen shot - iphone 5s - 2018-02-03 at 12 28 32

AlexRamey:
simulator screen shot - iphone 5s - 2018-02-03 at 12 28 47

AlexRamey:
simulator screen shot - iphone x - 2018-02-03 at 13 20 37

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -49,12 +77,29 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
         case .loading:
             break
         case .loaded(let loadedDevotions):
-            self.devotions = loadedDevotions
-            tableView.reloadData()
+            if self.devotions.count == 0 {
```
jdwitten, line 80: I think we should reload data every time there is an update to devotions in the state unless you have a reason not to?

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -49,12 +77,29 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
+                self.devotions = loadedDevotions
+                tableView.reloadData()
+            }
+        }
+    }
+    
+    func scrollToSelectedDevotion(animated: Bool) {
```
jdwitten, line 87: Nice dude I like this

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -49,12 +77,29 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
+        guard let selectedRow = devotions.index(where: { (devotion) -> Bool in
+            if let selectedDate = self.latestSelectedDate?.convertedDate(),
```
jdwitten, line 89: How does `self.latestSelectedDate` get set for the first time? I see that when you select a date on the calendar it gets set, but what date does it scroll to when this gets called in `viewDidLoad`?

jdwitten (commented):
Couple questions for you, I'll wait till you answer before approving but I like this man. The calendar looks straight cash

iOS Client/DevotionsController/MBDevotionsViewController.swift
AlexRamey: So I call this from two places: `viewDidAppear` and `didSelectDayView`. As you mentioned, `didSelectDayView` sets the `latestSelectedDate`.

The trick is, the calendar calls `didSelectDayView` with the current date when it loads, so it's ready to go in `viewDidAppear`. I think I avoided doing stuff in `viewDidLoad` b/c it was too early to scroll the table view since the frames weren't all setup yet (the scroll was slightly off). The library made me put those two lines in `viewDidLayoutSubviews` and I think it resizes itself a bit there.

iOS Client/DevotionsController/MBDevotionsViewController.swift
AlexRamey: if we were doing substate stuff, it'd be better. 

I can't remember if there was actually an issue though. I'll think about this more tomorrow.

I did feel bad about cutting out ReSwift. I basically started tracking `isRead` state internally (as well as firing actions to persist it to disk) and ignoring subsequent state updates here . . .

PR #55, opened 2018-02-22
dropped in calendar component for better devotions navigation

AlexRamey:


iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -49,12 +77,36 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
-            tableView.reloadData()
+            if self.devotions.count != loadedDevotions.count {
+                self.devotions = loadedDevotions
+                tableView.reloadData()
+            } else {
+                for (idx, loadedDevotion) in loadedDevotions.enumerated() {
+                    if self.devotions[idx] != loadedDevotion {
+                        self.devotions[idx] = loadedDevotion
+                        tableView.reloadRows(at: [IndexPath(row: idx, section: 0)], with: .automatic)
+                    }
```
jdwitten: should we batch up the indexes that need to be reloaded and then reload outside of the `for` loop? Probably a tad more efficient than calling `reloadRows` for each devotion that has changed.

iOS Client/DevotionsController/MBDevotionsViewController.swift
```diff
@@ -49,12 +77,36 @@ class MBDevotionsViewController: UIViewController, StoreSubscriber, UITableViewD
...
+                }
+            }
+        }
+    }
+    
+    func scrollToSelectedDevotion(animated: Bool) {
```
jdwitten, line 96: Nice I like it

iOS Client/Middleware/MBMiddleware.swift
```diff
@@ -24,4 +24,22 @@ struct MiddlewareFactory {
     }
     
     // Next Middleware . . .
+    static let saveDevotionMiddleware: Middleware<Any> = { dispatch, getState in
```
jdwitten, line 27: Yo this middle ware syntax kind of confuses me, but if I'm understanding correctly it will try to save the devotion and if it does, then it will continue to next middleware, if not then it will not pass on?

jdwitten (dismissed):
Alright my dude I think it looks good, just a small comment about reloading table view. I'll leave it up to you whether you think it is necessary to change. But feel free to merge whenever you're ready

iOS Client/DevotionsController/MBDevotionsViewController.swift
AlexRamey: ooooooo @jdwitten strong. I'll address this.

iOS Client/Middleware/MBMiddleware.swift
AlexRamey: that's right. If there's an error, it swallows it and doesn't even go to the reducers . . . which kind of haunts me a bit. B/c, if a user clicks a table view cell, and then we can't mark the devotion as read b/c an error occurs here, then they don't go anywhere, and they don't get an error. An error would be very unlikely . . . maybe worth dispatching a new action in the event of an error that triggers an alert view or something . . .

AlexRamey: I altered this to be best-effort. Saving as a pure side-effect. So, even if it fails, I print an error but let the UDF continue so the user doesn't perceive any issues.

They just might notice that the red dot doesn't go away

jdwitten (approved):

PR #56, opened 2018-02-27
ui improvements for Podcast detail

jdwitten:
Just worked on the UI for podcast detail. Next step will be to get the now playing bar above tab bar looking a little better


iOS Client/HTTPClient/MBClient.swift
```diff
@@ -52,6 +52,14 @@ class MBClient: NSObject {
             case .mockingCast: return "mockingcast"
             }
         }
+        
+        var title: String {
+            switch self {
+            case .pz: return "PZ Podcast"
+            case .mockingCast: return "The MockingCast"
+            case .mockingPulpit: return "Mocking Pulpit"
```
AlexRamey: Not completely sure, but I think we want `The Mockingpulpit` `The Mockingcast` and `PZ's Podcast` based on the titles that show up in iTunes (which seem to be derived from the Title field at the top of the XML returned by those 3 URLs . . . perhaps we can dynamically figure it out instead of hard-coding it . . . Not completely sure. Hard-coding isn't a major issue since the urls have the name as well, but It'd be stylish if we could rely on the data source)

iOS Client/PodcastsController/MBPodcastsViewController.swift
```diff
@@ -25,6 +25,7 @@ class MBPodcastsViewController: UIViewController, UITableViewDataSource, UITable
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+        self.navigationController?.navigationBar.isTranslucent = false
```
AlexRamey, line 28: does this belong in the Podcast's coordinator's start() function or the nav controller lazy property code? Curious about your thoughts. 

I've been altering the navigation Item of the current view controller in the view controller, which pretty much assumes knowledge of the containing nav controller.

This goes one step further, and actually alters a property on the parent nav controller's nav bar.

I actually think it's okay if it remains here after thinking about it, b/c it has to do with appearance and won't crash if someone were to drop this VC in a context where it is not embedded in a nav controller.

iOS Client/Models/MBPodcast.swift
```diff
@@ -29,4 +29,5 @@ struct Podcast: Detailable {
     let summary: String?
     let pubDate: Date
     let title: String?
+    let podTitle: String
```
AlexRamey: `feedName` `streamName` or `channelName`? podTitle seems a little ambiguous

iOS Client/Assets.xcassets/pause-bars.imageset/Contents.json
```diff
@@ -2,17 +2,15 @@
   "images" : [
     {
       "idiom" : "universal",
-      "filename" : "ic_pause.png",
+      "filename" : "pause-bars.png",
       "scale" : "1x"
     },
     {
       "idiom" : "universal",
```
AlexRamey, line 9: interesting, are we just rolling with one image for now?

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -20,19 +20,31 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
     @IBOutlet weak var stackView: UIStackView!
     @IBOutlet weak var playPauseButton: UIButton!
     @IBOutlet weak var durationLabel: UILabel!
+    @IBOutlet weak var totalDurationLabel: UILabel!
+    @IBOutlet weak var imageWidthConstraint: NSLayoutConstraint!
+    @IBOutlet weak var imageHeightConstraint: NSLayoutConstraint!
```
AlexRamey, line 25: strong

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -79,17 +117,27 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+           imageView.image = UIImage(named: imageName)
+            backgroundImageView.image = UIImage(named: imageName)
+        }
+        self.navigationItem.title = state.podcastsState.selectedPodcast?.podTitle
     }
     
     func updateCurrentDuration(current: Double, total: Double ) {
         totalDuration = total
-        durationLabel.text = "\(formatter.string(from: NSNumber(value: current)) ?? "0") / \(formatter.string(from: NSNumber(value: total)) ?? "0")"
+        timeFormatter?.allowedUnits = total >= Double(3600) ? [.hour, .minute, .second] : [.minute, .second]
```
AlexRamey, line 138: nice, so, if this is derived each time, can we delete line 67 in the configureFormatter function?

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -79,17 +117,27 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
     
     func newState(state: MBAppState) {
         switch state.podcastsState.player {
-        case .error, .initialized, .paused, .finished:
-            playPauseButton.setImage(#imageLiteral(resourceName: "play"), for: .normal)
+        case .initialized, .error, .paused, .finished:
+            playPauseButton.setImage(#imageLiteral(resourceName: "play-arrow"), for: .normal)
+            if playerState == .playing, playerState != .initialized { animateImage(size: CGSize(width: 150, height: 150), duration: 0.6)}
```
AlexRamey, line 122: if `playerState == .playing` then it follows that `playerState != .initialized` b/c `.playing != .initialized`. Thus, we can delete the second check.

AlexRamey, line 122: Can we pull these up as `let` constants on this VC?

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -79,17 +117,27 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
         case .playing:
-            playPauseButton.setImage(#imageLiteral(resourceName: "pause"), for: .normal)
+            playPauseButton.setImage(#imageLiteral(resourceName: "pause-bars"), for: .normal)
+            if playerState != .playing, playerState != .initialized { animateImage(size: CGSize(width: 200, height: 200), duration: 0.6)}
```
AlexRamey, line 125: why the check for initialized? It seems to me you're trying to catch toggles between playing and not playing to do some image animation

Going from `initialized` to `playing` seems like a good time to blow up the image. Btw, constants might be good here too. They can live in this file at the top. That'll put config in one place

iOS Client/PodcastsController/PodcastsCoordinator.swift
```diff
@@ -57,8 +57,11 @@ class PodcastsCoordinator: NSObject, Coordinator, StoreSubscriber, AVAudioPlayer
     }
     
     @objc func updateDuration(_ sender: Timer) {
-        if let totalDuration = player.currentItem?.duration.seconds {
-            podcastDetailViewController?.updateCurrentDuration(current: getCurrentDuration(), total: totalDuration)
+        let current = getCurrentDuration()
+        if let totalDuration = player.currentItem?.duration.seconds, !totalDuration.isNaN, !current.isNaN  {
```
AlexRamey, line 61: `isNaN` . . . very cool.

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -71,6 +90,25 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
         }
     }
     
+    func animateImage(size: CGSize, duration: CFTimeInterval) {
+        var shadowRect = imageView.bounds
+        var fromRect = CGRect(x: shadowRect.origin.x, y: shadowRect.origin.y, width: shadowRect.width + 10, height: shadowRect.height + 10)
+        shadowRect.size = CGSize(width: size.width + 10, height: size.height + 10)
+        let shadowAnimation = CABasicAnimation(keyPath: "shadowPath")
```
AlexRamey: not 100% if it'll work but try the `#keypath(UIImage.shadowPath)` pattern (not sure if that's the right path, maybe it's UIImage.layer.shadowPath or something )

AlexRamey: it will avoid stringly typed code trap

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -71,6 +90,25 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        shadowAnimation.fromValue = UIBezierPath(rect: fromRect).cgPath
+        shadowAnimation.toValue = UIBezierPath(rect: shadowRect).cgPath
+        shadowAnimation.isRemovedOnCompletion = false
+        shadowAnimation.fillMode = kCAFillModeForwards
+        shadowAnimation.duration = duration
+        shadowAnimation.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+        self.imageWidthConstraint.constant = size.width
+        self.imageHeightConstraint.constant = size.height
```
AlexRamey, line 105: strong, it's always a good day when you get to reference auto layout constraints in your code

iOS Client/PodcastsController/PodcastDetailViewController.swift
```diff
@@ -71,6 +90,25 @@ class PodcastDetailViewController: UIViewController, StoreSubscriber {
...
+        self.imageView.layer.add(shadowAnimation, forKey: "shadowPath")
```
AlexRamey: try to work in the `#keypath()` magic

AlexRamey (changes requested):
Very nice 👍 I pulled it down, really like the image growing and shrinking as you toggle play/pause.

I got a bunch of constraint printouts when I ran it locally on the iPhone X that seemed to be pointing to some issue we need to resolve. I was able to hit the debugger play button a few times to power through it and then the app worked.

iOS Client/HTTPClient/MBClient.swift
jdwitten: Yeah I'll change that to the strings for now and maybe if we get to it we can parse out the podcast name from the rss feed

iOS Client/PodcastsController/MBPodcastsViewController.swift
jdwitten: yeah this needs to be here because when coming back from podcast detail the nav bar needs to change from translucent to non translucent

iOS Client/Models/MBPodcast.swift
jdwitten: strong -- will change this

iOS Client/Assets.xcassets/pause-bars.imageset/Contents.json
jdwitten: The other image was too small and getting warped at large sizes so just found a new image. we can get some larger scales later if necessary

iOS Client/PodcastsController/PodcastDetailViewController.swift
jdwitten: Can't remember exactly why I check for initialized here. I think it has to do with when the view appears it looks funky if it is animating while pushing onto the screen.

iOS Client/PodcastsController/PodcastDetailViewController.swift
jdwitten: I'll check it out not exactly sure how `#keypath` works but I'll give it a shot

iOS Client/PodcastsController/PodcastDetailViewController.swift
jdwitten: yeah that is why ^^^

AlexRamey: ok cool

AlexRamey (approved):
glad #keyPath worked out, it is very classy

PR #57, opened 2018-03-20
only auto-refresh after subscribing to the state (so articles have al…

AlexRamey:
…ready loaded) on first appearance

solves #51


iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -36,7 +36,7 @@ class MBPodcastsStore {
             for (indx, response) in responses.enumerated() {
                 if case .fulfilled(let newCasts) = response {
                     let displayCasts = newCasts.flatMap { podcast -> Podcast? in
-                        guard let dateString = podcast.pubDate, let date = self.dateFormatter.date(from: dateString) else {
```
AlexRamey, line 39: this was filtering out all mocking pulpit podcasts b/c their dates had EDT instead of -0500 at the end

jdwitten (approved):

PR #62, opened 2018-03-24
Feature/now playing ui improvements

jdwitten:


iOS Client/Utilities/UIView.swift
```diff
@@ -0,0 +1,25 @@
...
+//  iOS Client
+//
+//  Created by Jonathan Witten on 3/24/18.
+//  Copyright © 2018 Mockingbird. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+protocol LoadableNibByClassName { }
```
AlexRamey, line 12: clever

iOS Client/TabController/MBTabBarController.swift
```diff
@@ -71,15 +67,18 @@ class MBTabBarController: UITabBarController, StoreSubscriber {
     }
     
     func newState(state: MBAppState) {
-        
+        if let podImageName = state.podcastsState.selectedPodcast?.image {
+            playPauseView.image.image = UIImage(named: podImageName)
```
AlexRamey, line 71: It seems like we could check if the image is already set, then don't do any work.

AlexRamey: would also need to track the old state to make sure we overwrite incorrect images

AlexRamey (approved):

PR #67, opened 2018-04-10
Add filtering capability for podcasts

jdwitten:


iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
```diff
@@ -0,0 +1,35 @@
...
+class PodcastFilterTableViewCell: UITableViewCell {
+
+    @IBOutlet weak var toggleSwitch: UISwitch!
+    @IBOutlet weak var titleLabel: UILabel!
+    @IBOutlet weak var podcastImage: UIImageView!
+    
+    var podcast: PodcastStream?
+    
+    static var reuseIdentifier: String = "PodcastFilterTableViewCell"
+    func configure(image: UIImage? = nil, podcast: PodcastStream, on: Bool = false) {
```
AlexRamey: Something I read online reminded me that this pattern is a violation of MVC. By passing the PodcastStream model to the TableViewCell, the view is talking directly to the model.

It does provide a nice way to get this code out of the VC. Idk. In articles, I just set everything on the view from the VC for now. Curious to hear your thoughts.

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
```diff
@@ -0,0 +1,35 @@
...
+
+class PodcastFilterTableViewCell: UITableViewCell {
+
+    @IBOutlet weak var toggleSwitch: UISwitch!
+    @IBOutlet weak var titleLabel: UILabel!
+    @IBOutlet weak var podcastImage: UIImageView!
+    
+    var podcast: PodcastStream?
+    
+    static var reuseIdentifier: String = "PodcastFilterTableViewCell"
```
AlexRamey: interesting, coupling the reuse identifier with the cell class. I typically don't see this pattern. Since tableViews register a certain cell class (or nib) for a certain reuse identifier, it seems like you want control over what to make the reuse identifier. The cell usually doesn't prescribe it. Kind of makes sense in case multiple cells happened to have a namespace collision with their reuse identifiers. 

From a code reusability perspective, you may be happier down the road if this cell container is a little bit more generic and dumb without bundling in these details. 

As stylish as this is, I think its a bit dangerous. Of course, not going to fret about it too much for this project.

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
```diff
@@ -0,0 +1,35 @@
...
+    func configure(image: UIImage? = nil, podcast: PodcastStream, on: Bool = false) {
+        toggleSwitch.isOn = on
+        titleLabel.text = podcast.title
+        podcastImage.image = image
+        self.podcast = podcast
+        podcastImage.layer.cornerRadius = 10
```
AlexRamey: can this be done just once in `awakeFromNib`?

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
```diff
@@ -0,0 +1,35 @@
...
+        self.selectionStyle = .none
+    }
+    @IBAction func toggled(_ sender: UISwitch) {
+        guard let podcast = podcast else {
+            return
+        }
+        MBStore.sharedStore.dispatch(TogglePodcastFilter(podcastStream: podcast, visible: toggleSwitch.isOn))
```
AlexRamey: gotta love UDF

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import UIKit
+
+class PodcastFilterTableViewCell: UITableViewCell {
+
+    @IBOutlet weak var toggleSwitch: UISwitch!
+    @IBOutlet weak var titleLabel: UILabel!
+    @IBOutlet weak var podcastImage: UIImageView!
+    
+    var podcast: PodcastStream?
```
AlexRamey: Would `stream` be a more descriptive variable name?

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
```diff
@@ -0,0 +1,35 @@
...
+    
+    static var reuseIdentifier: String = "PodcastFilterTableViewCell"
+    func configure(image: UIImage? = nil, podcast: PodcastStream, on: Bool = false) {
+        toggleSwitch.isOn = on
+        titleLabel.text = podcast.title
+        podcastImage.image = image
+        self.podcast = podcast
+        podcastImage.layer.cornerRadius = 10
+        self.selectionStyle = .none
```
AlexRamey: This is very cool. I typically disable user interaction but this is better.

Again, it can probably be set once in `awakeFromNib`

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
AlexRamey: Just a thought, but, if we're toggling something, and the state is tracking all of this stuff, then wouldn't the state know just to toggle it

`new_state = !old_state`? Why pass visible param?

iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -0,0 +1,72 @@
...
+
+class PodcastsFilterViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, StoreSubscriber {
+    
+    @IBOutlet weak var tableView: UITableView!
+    var visibleStreams: Set<PodcastStream> = Set<PodcastStream>()
+    var streams: [PodcastStream] = [.pz, .mockingCast, .mockingPulpit]
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        tableView.register(UINib(nibName: PodcastFilterTableViewCell.reuseIdentifier, bundle: nil), forCellReuseIdentifier: PodcastFilterTableViewCell.reuseIdentifier)
```
AlexRamey: My concern with making the reuse identifier match the nib name is the two sources of truth. We're now in a situation where if either the reuse identifier changes or the name of the view file changes then this code will break.

AlexRamey: if we hard-code the nib name, then at least changing the reuse identifier doesn't break the code.

iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -0,0 +1,72 @@
...
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.rowHeight = UITableViewAutomaticDimension
+        tableView.estimatedRowHeight = 110
+        tableView.tableFooterView = nil
```
AlexRamey, line 27: is this necessary?

iOS Client/PodcastsController/PodcastTableViewCell.swift
```diff
@@ -11,6 +11,7 @@ import UIKit
 class PodcastTableViewCell: UITableViewCell {
```
AlexRamey, line 13: whitespace

iOS Client/State/MBState.swift
```diff
@@ -68,12 +68,14 @@ protocol PodcastsState {
...
     var selectedPodcast: Podcast? { get set }
     var player: PlayerState { get set }
+    var visiblePodcasts: Set<PodcastStream> { get set }
 }
 
 struct MBPodcastsState: PodcastsState {
     var podcasts: Loaded<[Podcast]> = .initial
     var selectedPodcast: Podcast?
     var player: PlayerState = .initialized
+    var visiblePodcasts: Set<PodcastStream> = Set<PodcastStream>([.pz, .mockingPulpit, .mockingCast]) // start with all podcasts visible
```
AlexRamey: flagged

iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -0,0 +1,72 @@
...
+
+import Foundation
+import ReSwift
+import UIKit
+
+class PodcastsFilterViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, StoreSubscriber {
+    
+    @IBOutlet weak var tableView: UITableView!
+    var visibleStreams: Set<PodcastStream> = Set<PodcastStream>()
+    var streams: [PodcastStream] = [.pz, .mockingCast, .mockingPulpit]
```
AlexRamey: We are falling into another subtle trap here: two sources of truth.

This list exists here and in the state initialization (which I flagged). Perhaps this information should come from the state upon newState call or something

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -27,7 +27,7 @@ class MBPodcastsStore {
     }
     
     func syncPodcasts() -> Promise<[Podcast]> {
-        let streams: [MBClient.PodcastStream] = [.pz, .mockingPulpit, .mockingCast]
+        let streams: [PodcastStream] = [.pz, .mockingPulpit, .mockingCast]
```
AlexRamey, line 30: We could pass these streams in a parameter to achieve one source of truth perhaps.

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -57,3 +57,25 @@ class MBPodcastsStore {
         }
     }
 }
+
+public enum PodcastStream: String {
```
AlexRamey: this should live with the models, since it is referenced by the Podcast model. 

These models are the center of the Onion. If we define it here, at this outer layer, then it violates the Dependency Inversion Principle of Clean Code Architecture for other models to know about this one.

Again, I'm not going to fret over this.

AlexRamey (dismissed):
I did a lot of rambling here.

Feel free to merge or address the comments as you see fit.

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
jdwitten: Yeah I see what you're saying. It's just a matter of whether to list out properties one by one in the constructor or just have one for the entire model. I'm ok with changing it

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
jdwitten: very true I think it would

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
jdwitten: nah needs to be done here because this cell can be reused so it needs all of its properties to be reset on `configure`

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
jdwitten: This might be able to be set in `awakeFromNib` because no matter the content it still won't be selectable

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
jdwitten: Ummm good point I think that might work

iOS Client/PodcastsController/PodcastFilterViewController.swift
jdwitten: So I don't think this is necessarily a violation of two sources of truth. We are deriving visibleStreams straight from the state in `newState` see `state.podcastsState.visiblePodcasts` and then using it as the data source for the table view. This is similar to storing the list of accounts straight from the state so when the table view reloads it has data to reference.

iOS Client/PodcastsController/PodcastFilterViewController.swift
jdwitten: Yeah it prevents the table view from having trailing empty cells

iOS Client/Store/MBPodcastsStore.swift
jdwitten: Interesting so where are you suggesting to put this?

iOS Client/PodcastsController/PodcastFilterViewController.swift
jdwitten: Oh I think I see what you're saying. Derive the actual list of podcasts from state. Yeah that is something that I've been thinking about, we are hardcoding the list but in my opinion that's a pretty safe assumption because we aren't really deriving that list from anywhere except requirements from DZ which would take an app update to change anyways

jdwitten: I think I'm going to leave this as a possible future refactor because it would take some rethinking on how we are managing the streams throughout the whole podcast flow if we were to derive the list of streams from a central data source (which honestly as I'm thinking more about it is probably the right way to go)

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
AlexRamey: Up to you. The VC code could also just set properties directly

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
AlexRamey: does `podcastImage.layer` get reset when `podcastImage.image` changes? I'm not 100% sure how that works

iOS Client/PodcastsController/PodcastFilterViewController.swift
AlexRamey: very strong, good to know

iOS Client/PodcastsController/PodcastFilterTableViewCell.swift
jdwitten: Yeah honestly I don't have a strong opinion about reuse identifiers. I think on Synch we keep them with the cell class like I have here, but not too positive about the ramifications of that. I'll move them to properties on the VC to decouple

iOS Client/Store/MBPodcastsStore.swift
AlexRamey: I'd say `Models` folder, maybe in its own file. 

On a side note, our DTOs should really not be in Models, but with the client code probably. Again, not a huge deal.

jdwitten:
Yo I think I need another review because I pushed some more commits. I addressed many of your comments, but still need improvements on the whole Onion principle for podcast streams. I'll make an issue for that and we can address it if we get to it

iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -14,11 +14,12 @@ class PodcastsFilterViewController: UIViewController, UITableViewDataSource, UIT
...
     @IBOutlet weak var tableView: UITableView!
     var visibleStreams: Set<PodcastStream> = Set<PodcastStream>()
-    var streams: [PodcastStream] = [.pz, .mockingCast, .mockingPulpit]
+    var streams: [PodcastStream] = []
+    let filterReuseIdentifier: String = "PodcastFilterTableViewCell"
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        tableView.register(UINib(nibName: PodcastFilterTableViewCell.reuseIdentifier, bundle: nil), forCellReuseIdentifier: PodcastFilterTableViewCell.reuseIdentifier)
+        tableView.register(UINib(nibName: "PodcastFilterTableViewCell", bundle: nil), forCellReuseIdentifier: filterReuseIdentifier)
```
AlexRamey: nice

iOS Client/Reducers/MBAppReducer.swift
```diff
@@ -98,12 +98,14 @@ func podcastsReducer(action: Action, state: PodcastsState?) -> PodcastsState {
...
+        if state?.visibleStreams.contains(action.podcastStream) ?? false {
+            nextState.visibleStreams.remove(action.podcastStream)
         } else {
-            nextState.visiblePodcasts.remove(action.podcastStream)
+            nextState.visibleStreams.insert(action.podcastStream)
         }
-        
+    case let action as SetPodcastStreams:
+        nextState.streams = action.streams
+        nextState.visibleStreams = Set<PodcastStream>(action.streams.map { $0 })
```
AlexRamey: is the call to map necessary?

iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -52,30 +52,9 @@ class MBPodcastsStore {
                     podcasts.append(contentsOf: displayCasts)
                 }
             }
+            MBStore.sharedStore.dispatch(SetPodcastStreams(streams: [.pz, .mockingPulpit, .mockingCast]))
```
AlexRamey, line 55: I think it'd be fine to hard-code this into the initial state; if you still want to dispatch this action tho, I think u can reuse streams defined on line 30 of this file instead of making new array

AlexRamey (approved):

PR #102, opened 2018-08-05
Add support for option to just show downloaded podcasts

jdwitten:


iOS Client/Store/MBPodcastsStore.swift
```diff
@@ -133,18 +133,26 @@ class MBPodcastsStore: PodcastsRepository {
...
-        return visibilitySettings.keys.compactMap { PodcastStream(rawValue: $0)}.filter { visibilitySettings[$0.rawValue] ?? false}
+        return visibilitySettings.keys.compactMap {
+            if let stream = PodcastStream(rawValue: $0) {
+                return PodcastFilterOption.stream(stream)
+            } else if $0 == PodcastFilterOption.downloaded.key {
+                return PodcastFilterOption.downloaded
+            } else {
+                return nil
+            }
+        }.filter { visibilitySettings[$0.key] ?? false}
```
AlexRamey: is this filter necessary?

AlexRamey: it would make more sense to filter first I think

iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -10,11 +10,21 @@ import UIKit
...
+    var visibleOptions: [PodcastFilterOption] = []
     var repository: PodcastsRepository!
     
     let filterReuseIdentifier: String = "PodcastFilterTableViewCell"
     
+    var shouldHideStreamOptions: Bool {
+        return visibleOptions.contains(where: {
+            switch $0 {
+            case .downloaded: return true
+            case .stream: return false
```
AlexRamey: so, if downloaded is toggled on, then we hide the other choices . . . sounds reasonable

iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -10,11 +10,21 @@ import UIKit
 
 class PodcastsFilterViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, PodcastFilterDelegate {
     @IBOutlet weak var tableView: UITableView!
-    var streams: [PodcastStream] = []
+    var filterOptions: [PodcastFilterOption] = []
+    var visibleOptions: [PodcastFilterOption] = []
```
AlexRamey: this is a confusing name. Based on the code, this seems to be `enabledOptions`. Visible options implies these are the options present in the table view

iOS Client/PodcastsController/MBPodcastsViewController.swift
```diff
@@ -67,9 +67,22 @@ class MBPodcastsViewController: UIViewController, UITableViewDataSource, UITable
...
+            else { return false }
+        })
         self.visiblePodcasts = self.podcasts.filter({ (podcast) -> Bool in
-            visibleStreams.contains(podcast.feed)
+            if showDownloadedPodcasts {
+                guard savedPodcastTitles.contains(podcast.title ?? "") else {
+                    return false
+                }
+            }
+            return visibleStreams.contains(podcast.feed)
```
AlexRamey: this still respects the filter settings. 

I'm not sure we want to do that if we're going to be hiding those filter options

iOS Client/PodcastsController/PodcastFilterViewController.swift
jdwitten: yeah I'll change to enabledOptions, better name

iOS Client/Store/MBPodcastsStore.swift
jdwitten: Actually I don't think we really even need the filter at all here. If the item in the array makes it past the compactMap then of course it's going to be in the visibilitySettings

iOS Client/PodcastsController/MBPodcastsViewController.swift
jdwitten: yeah good point, I'll change

AlexRamey (approved):
lgtm

PR #104, opened 2018-08-05
create a podcast description screen with descriptions for each podcas…

jdwitten:
…t stream


iOS Client/PodcastsController/PodcastFilterViewController.swift
```diff
@@ -24,6 +29,7 @@ class PodcastsFilterViewController: UIViewController, UITableViewDataSource, UIT
         tableView.estimatedRowHeight = 110
         tableView.tableFooterView = nil
         navigationItem.title = "Filter"
+        navigationItem.rightBarButtonItem = UIBarButtonItem(image: UIImage(named: "info"), style: .done, target: self, action: #selector(PodcastsFilterViewController.viewInfo))
```
AlexRamey: I think there's a way to do this without bringing in an image

AlexRamey: see https://github.com/AlexRamey/CharlottesvilleArtfinder-iOS/blob/master/Charlottesville%20Artfinder/ARTDirectoryViewController.m

iOS Client/PodcastsController/PodcastInfoViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+            let image = UIImage(named: stream.imageName) {
+            cell.configure(image: image, name: stream.title, description: stream.description)
+            return cell
+        } else {
+            return UITableViewCell()
+        }
+    }
+    
+    @objc func dismissInfo(_ sender: Any) {
+        handler?.dismissInfo()
```
AlexRamey: I don't think this indirection is necessary. You can just do self.dismiss()

AlexRamey: I don't think this indirection is necessary. You can just do self.dismiss here

iOS Client/PodcastsController/PodcastInfoTableViewCell.swift
```diff
@@ -0,0 +1,27 @@
...
+    @IBOutlet weak var titleLabel: UILabel!
+    
+    func configure(image: UIImage, name: String, description: String) {
+        self.titleLabel.text = description
+        self.podcastImage.image = image
+        self.podcastName.text = name
+        self.titleLabel.font = UIFont(name: "IowanOldStyle-Roman", size: 18)
+        self.podcastName.font = UIFont(name: "IowanOldStyle-Roman", size: 14)
+        self.podcastName.textColor = UIColor.gray
+        self.selectionStyle = .none
```
AlexRamey: these bottom 4 rows don't need to be set every time configure is called. They can be moved up into an awakeFromNib function

iOS Client/PodcastsController/PodcastInfoTableViewCell.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+//  Created by Jonathan Witten on 8/4/18.
+//  Copyright © 2018 Mockingbird. All rights reserved.
+//
+
+import UIKit
+
+class PodcastInfoTableViewCell: UITableViewCell {
+
+
```
AlexRamey, line 13: whitespace

AlexRamey (changes requested):
let's try to use the system icon instead of bringing in an image: https://github.com/AlexRamey/CharlottesvilleArtfinder-iOS/blob/master/Charlottesville%20Artfinder/ARTDirectoryViewController.m

AlexRamey (approved):

